PR #19, opened 2018-07-15
make electrs listen port and bitcoin rpc url configurable

greenaddress:
I tested this locally and seems to work ok. I wanted to check if there was interest and if there's a preference on how to go about it.

This PR is meant to add configurability for the tcp listening port (default to 50001 for mainnet) and for the bitcoin rpc ip:port.

It attempts to avoid repeating the default ports around the code and it defaults to correct ports for mainnet/testnet - i.e. the configurability is meant to be optional.

I'll make updates to the PR based on feedback/interest, results of additional testing, etc

### DIFF
--- a/examples/index.rs
+++ b/examples/index.rs
@@ -18,6 +18,7 @@ fn run() -> Result<()> {
 
     let daemon = Daemon::new(
         &config.daemon_dir,
+        &config.daemon_rpc_url,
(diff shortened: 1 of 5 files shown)


romanz:
Thanks a lot for the PR!
Merged with a few small fixes :)

Issue #33, opened 2018-07-19
Travis build fails due to '/usr/bin/ld: final link failed: No space left on device'


romanz:
https://travis-ci.com/romanz/electrs/builds/79488852#L513

PR #35, opened 2018-07-20
fix travis out of memory issue

greenaddress:
Hopefully fixes https://github.com/romanz/electrs/issues/33 and ok as a workaround


romanz:
Thanks a lot for the fix!

PR #51, opened 2018-08-21
remove unused size_on_disk

greenaddress:
Not sure if this was intentional, if so please close this PR.

If it wasn't intentional this removes size_on_disk as a field in the BlockchainInfo struct since it is not used anywhere. As per usage.md changes, this does affect logging - since it only affects bitcoind logging it may be ok as user can still log getblockchaininfo via bitcoin-cli.


romanz:
Thanks for the PR and sorry for the delayed response...
This field was added only for debugging purposes, so it can be safely removed.

PR #105, opened 2018-12-14
Electrum protocol v1.4

dagurval:
This adds the changes required to be compatible protocol 1.4 as defined by electrumx
https://electrumx.readthedocs.io/en/latest/protocol-changes.html#version-1-4

Lightly tested by comparing output from electrumx and by running the master branch of [Electron Cash](https://github.com/Electron-Cash/Electron-Cash)

This is the first time I'm trying out Rust, so nits are more then welcome.

PR #108, opened 2018-12-19
tx_id -> tx_hash in transaction.id_from_pos

dagurval:
Bug found in review of #105 

https://electrumx.readthedocs.io/en/latest/protocol-methods.html#blockchain-transaction-id-from-pos


romanz:
Thanks!

PR #116, opened 2019-02-22
Resolve hostname for arguments passing an address

dagurval:
This allows hostname to be passed, rather than only IP addresses.

My use case for this is having the bitcoin node running on a separate docker instance. The name of the other docker instance resolves to its docker network IP.


dagurval:
A simple way to test this is to pass `--daemon-rpc-addr=localhost:8332`. It should be visible in the log that this resolves to `127.0.0.1:8332`.

PR #127, opened 2019-04-01
[mempool] Fix bug on removing entries

dagurval:
Fixes issue where old mempool entires were only removed if new ones were
also added.

Diff is best viewed ignoring whitespace (add [`?w=1`](https://github.com/romanz/electrs/pull/127/files?w=1))


romanz:
Thanks!

PR #144, opened 2019-05-29
Use atomic u64 instead of Mutex for counter

Kixunil:
This change switches from `Mutex<u64>` to `AtomicU64`, which is faster
and somewhat more elegant to write. It keeps the `Counter` type for
backwards compatibility.

The downside is that it doesn't work on as many platforms and requires
fairly recent compiler.

### DIFF
--- a/src/daemon.rs
+++ b/src/daemon.rs
@@ -13,7 +13,7 @@ use std::collections::{HashMap, HashSet};
 use std::io::{BufRead, BufReader, Lines, Write};
 use std::net::{SocketAddr, TcpStream};
 use std::path::PathBuf;
-use std::sync::{Arc, Mutex};
+use std::sync::{Arc, Mutex, atomic::{AtomicU64, Ordering}};
 use std::time::Duration;
 
 use crate::errors::*;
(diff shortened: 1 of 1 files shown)


Kixunil:
See [the documentation for `std::sync::atomic`](https://doc.rust-lang.org/std/sync/atomic/index.html) for more details about atomics.

romanz:
Thanks a lot for the PR!

The only question that I have is whether the new code runs on ARM-based devices (e.g. Raspberry Pi)?
/cc @sangaman @duckartes @jimmyongit

Stadicus:
I just compiled https://github.com/romanz/electrs/pull/144/commits/2a835d94ef8e99a9b53a2952936c69320171429b on a RockPro64 aarch64 board. No issues, runs fine without further testing.

```
$ clang --version
clang version 3.8.1-24 (tags/RELEASE_381/final)
Target: aarch64-unknown-linux-gnu
Thread model: posix
InstalledDir: /usr/bin
```
One warning during compilation, probably not relevant:
```
   Compiling librocksdb-sys v5.18.3
warning: redundant linker flag specified for library `stdc++`
```

romanz:
Unfortunately, it doesn't compile on Rust 1.32 (which is the current Debian version).
Will merge it after Debian updates to a more recent version.

Kixunil:
I was thinking that maybe a feature flag would help. It'd help with the other unsupported architectures as well. Thanks to the way `Counter` is implemented it'd be easy to do.

What do you think?

romanz:
Using a feature flag sounds great.
Could you please update the PR?

romanz:
Many thanks, I have added a feature flag for "latest" Rust support :)

Kixunil:
You are welcome. I wanted to do it myself but I was quite busy lately.

PR #145, opened 2019-05-29
Bunch of Vec<Row> translated to Iterator<Item=Row>

Kixunil:
There were several places where `Vec<Row>` was created just to iterate
it later and throw away, thus allocating memory unnecessarily. This
translates those `Vec`s to `Iterator`s in order to improve the
efficiency.

**Note** I didin't test this yet as my node is syncing, but the changes were pretty straightforward.

### DIFF
--- a/src/app.rs
+++ b/src/app.rs
@@ -27,7 +27,7 @@ impl App {
         }))
     }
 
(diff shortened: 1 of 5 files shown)


romanz:
Many thanks, much appreciated!
Will test soon :)

PR #153, opened 2019-07-04
Add cache for block transaction IDs

dagurval:
On cache hit, this improves performance of `blockchain.transaction.get_merkle` and thus wallet synchronization

### DIFF
--- a/src/cache.rs
+++ b/src/cache.rs
@@ -44,3 +44,54 @@ impl BlockTxIDsCache {
         Ok(txids)
     }
 }
+
+#[cfg(test)]
+mod tests {
+    use super::*;
+    use bitcoin_hashes::Hash;
+
+    fn gen_hash(seed: u8) -> Sha256dHash {
+        let bytes: Vec<u8> = (seed..seed + 32).collect();
+        Sha256dHash::hash(&bytes[..])
+    }
+
+    #[test]
+    fn test_cache_hit_and_miss() {
+        let block1 = gen_hash(1);
+        let block2 = gen_hash(2);
+        let block3 = gen_hash(3);
+        let txids = vec![gen_hash(4), gen_hash(5)];
+
+        let misses: Mutex<usize> = Mutex::new(0);
+        let miss_func = || {
+            *misses.lock().unwrap() += 1;
+            Ok(txids.clone())
+        };
+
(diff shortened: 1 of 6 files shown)


src/cache.rs
```diff
@@ -0,0 +1,46 @@
...
+            self.hits.inc();
+            return Ok(txids.clone());
+        }
+
+        self.misses.inc();
+        let txids = load_txids_func()?;
+        self.map.lock().unwrap().put(*blockhash, txids.clone());
+        Ok(txids)
+    }
+}
```
romanz: Could you please add a few tests for this module?

src/cache.rs
```diff
@@ -0,0 +1,46 @@
...
+        &self,
+        blockhash: &Sha256dHash,
+        load_txids_func: F,
+    ) -> Result<Vec<Sha256dHash>>
+    where
+        F: FnOnce() -> Result<Vec<Sha256dHash>>,
+    {
+        if let Some(txids) = self.map.lock().unwrap().get(blockhash) {
+            self.hits.inc();
+            return Ok(txids.clone());
```
romanz: Is it possible to avoid the `.clone()` here, by returning a `Result<&[Sha256dHash]>`?

src/cache.rs
dagurval: Test case added

src/cache.rs
dagurval: I'm not sure how. `LruCache::get` returns a reference.

dagurval: Maybe a way to do this would be to instead of returning the transaction list for processing, pass a callback for processing the transaction list in place, for example:
```
let result = self.app.daemon().withblocktxids(&header_entry.hash(), |txids| {
    let pos = txids
       .iter()
       .position(|txid| txid == tx_hash)
       .chain_err(|| format!("missing txid {}", tx_hash))?;
   let (branch, _root) = create_merkle_branch_and_root(txids, pos);
   Ok((branch, pos))
}
```
The trade-off is keeping a lock on the cache for a longer time. Also probably better to do in a separate PR.

romanz: > Maybe a way to do this would be to instead of returning the transaction list for processing, pass a callback for processing the transaction list in place

Good solution!

> Also probably better to do in a separate PR.

Agreed :)

PR #161, opened 2019-08-12
Switch to random replacement cache with capacity in megabytes

dagurval:
The current cache capacity is set as number of entries, rather than size in memory. Currently the smallest transaction on the BTC blockchain is 0.062kb, while the largest is 999.65.

So with the current cache default of 10000 entries, the cache usage could in theory be anywhere between 620kb to 10gb (although not enough transactions of that size exist in reality).

This PR implements a cache with capacity set in bytes, rather than number of entries. The cache uses random replacement over LRU. As implemented, the calculation of used bytes is not exact. The size of the cache key is not included in the usage, and also the overhead of the data structure for the cache entry is not included. But it is more predictable than what is currently implemented.

Rationale for using random replacement (RR) over LRU is to make the caching strategy non-deterministic. If the server is public, an attacker can when LRU is used predict what get evicted from cache, thus be able to intentionally cause cache misses. Using RR mitigates this attack.


src/config.rs
```diff
@@ -42,6 +42,10 @@ fn str_to_socketaddr(address: &str, what: &str) -> SocketAddr {
         .unwrap_or_else(|| panic!("no address found for {}", address))
 }
 
+fn mb_to_bytes(mb: usize) -> usize {
+    mb * 1000
```
dagurval: This conversion is thousandfold wrong

dagurval:
Pushed two commits. 1) Fix MB to byte conversion, 2) Don't fail if a cache entry would be too big for the cache.

romanz:
Sorry for the late response, was quite busy in the last weeks.
Will review this PR in the following days.

romanz:
It took me a while (sorry for the delay) but I think that we can adapt the existing cache with "byte-size" capacity.
In general, it's not recommended to run `electrs` on a publicly exposed port, and doing so will require more work (as was done for https://github.com/Blockstream/electrs fork).
Please take a look at https://github.com/romanz/electrs/pull/172 - would it work for your use-case?

dagurval:
> Would it work for your use-case?

The use case I’m aiming for is public servers. Your PR works though, it’s simple enough for me to swap out LRU with RR.

Closing in favour of #172

PR #166, opened 2019-09-06
Use configure_me instead of clap

### LINKED ISSUE #151: Use configure_me instead of clap?
Electrs now uses `clap` and command line for configuring the server. This has two significant downsides: there's lot of arguments that need to be passed and it's clunky to work with. The other issue is that by passing cookie (password) as argument, it's visible to other processes - including processes of other users.

In order to resolve this, I suggest to swap `clap` for [`configure_me`](https://crates.io/crates/configure_me), as that automatically generates the code for both argument parsing, env vars and loading of config files. The downside is it doesn't have all the features of `clap`. It worked well so far for my projects but there are still some things that need polishing.

As `configure_me` is my crate, I'm offering to implement this, since I know it well. Further, I offer to implement any feature that's missing in `configure_me` that's needed for `electrs`. The nice thing is that the generated code will create config struct exactly as the one used today.

I opened the issue first in order to not waste time if this idea is rejected.

What do you think?

Kixunil:
Since use of configuration files is both more secure and more
convenient and clap doesn't support config files, this switches to
configure_me, which supports config files, env vars and also generating
man pages.

Closes #151

This is WIP, I just wanted to show what I have before I go AFK for several hours. I tested it only very lightly. This also includes man page generation. :)

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -1,18 +1,114 @@
 use bitcoin::network::constants::Network;
-use clap::{App, Arg};
 use dirs::home_dir;
 use num_cpus;
 use std::fs;
+use std::fmt;
 use std::net::SocketAddr;
 use std::net::ToSocketAddrs;
-use std::path::{Path, PathBuf};
+use std::path::PathBuf;
 use std::sync::Arc;
+use std::str::FromStr;
+use std::ffi::{OsStr, OsString};
+use std::convert::TryInto;
 use stderrlog;
 
 use crate::daemon::CookieGetter;
 use crate::errors::*;
 
-const DEFAULT_SERVER_ADDRESS: &str = "127.0.0.1"; // by default, serve on IPv4 localhost
+const DEFAULT_SERVER_ADDRESS: [u8; 4] = [127, 0, 0, 1]; // by default, serve on IPv4 localhost
+
+mod internal {
+    #![allow(unused)]
+
+    include!(concat!(env!("OUT_DIR"), "/configure_me_config.rs"));
+}
+
+pub enum AddressError {
+    InvalidUtf8(OsString),
+    ResolvError { addr: String, err: std::io::Error },
+    NoAddrError(String),
+}
+
+impl fmt::Display for AddressError {
+    fn fmt(&self, f: &mut fmt::Formatter) -> fmt::Result {
+        match self {
+            AddressError::InvalidUtf8(val) => write!(f, "{:?} isn't a valid UTF-8 sequence", val),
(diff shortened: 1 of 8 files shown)


Kixunil:
Fixed `configure_me_codegen` line in `Cargo.toml` with force push as I feel such detail would needlessly clutter git history. I hope you don't mind.

src/config.rs
```diff
@@ -34,208 +130,89 @@ pub struct Config {
...
+        let network_name = match config.network {
+            Network::Bitcoin => "bitcoin",
+            Network::Testnet => "testnet",
+            Network::Regtest => "regtest",
         };
-        let db_dir = Path::new(m.value_of("db_dir").unwrap_or("./db"));
-        let db_path = db_dir.join(network_name);
 
-        let default_daemon_port = match network_type {
+        config.db_dir.push(network_name);
```
Kixunil: Damn, I just noticed this will change the name of the dir from `mainnet` to `bitcoin` this is pretty big backwards-incompatibility - going to fix it.

src/config.rs
```diff
@@ -1,18 +1,114 @@
...
+        match self {
+            AddressError::InvalidUtf8(val) => write!(f, "{:?} isn't a valid UTF-8 sequence", val),
+            AddressError::ResolvError { addr, err } => write!(f, "Failed to resolve address {}: {}", addr, err),
+            AddressError::NoAddrError(addr) => write!(f, "No address found for {}", addr),
+        }
+    }
+}
+
+
+#[derive(Deserialize)]
```
Kixunil: Oh, this doesn't resolve the address. I think it is a bad idea to try to resolve the address when parsing/deserializing - leads to confusing code. We still want a newtype to have a better description, but we should resolve instead of converting.

Kixunil:
I'm pretty happy with this now and don't see any important outstanding issues. I hope to test it even better tomorrow just to be sure.
Would you like to see any other change? Do you see any other issue?

romanz:
Looks great, much appreciated!
The only (non-critical) issue are:
 - formatting: Rust 1.37.0 (latest stable) seem to prefer the [following](https://github.com/romanz/electrs/commit/f7ad9419ed6f1e8358ed443f322da99110cc7c2b) style. WDYT?
 - release notes: I've added a few lines [here](https://github.com/romanz/electrs/commit/43e0bd31b15c25746a72d96ad0a8c57bc4949ad7). Should we elaborate more?

Kixunil:
I don't care too much about formatting, looks good. Release notes seem to be fine.

romanz:
Cool, thanks again for all the work!

PR #175, opened 2019-09-28
Support for user-specified onfiguration files/dirs

Kixunil:
This commit upgrades `configure_me` and adds support for `--conf` and
`--conf-dir` parameters, which load the configuration from the specified
place.


romanz:
Thanks!

Issue #168, opened 2019-09-07
Add integration tests


romanz:
e.g. run containerized `bitcoind -regtest`, and test that indexing & querying works a few transactions.
Links:
 - https://github.com/coblox/docker-bitcoin-core
 - https://github.com/rust-bitcoin/rust-bitcoincore-rpc

romanz:
Added here:
https://github.com/romanz/electrs/blob/master/tests/run.sh

PR #185, opened 2019-10-26
Replace PerodicUpdate with scripthash notifications

dagurval:
This removes the expensive `PeriodicUpdate` call from the application loop.

`PeriodicUpdate` would iterate through every subscription of every connected client and re-calculate its status to look for differences.

The new code calculates which subscription need update and notifies each RPC connection. If, and only if, a client is subscribed will it re-calculate its status.

To test performance, I have written [this benchmark script](https://github.com/BitcoinUnlimited/ElectrsCash/blob/master/contrib/bench/server-compare.js).

***Before this PR - DEBUG***
```
Connecting to  [ 60001, 'localhost', 'tcp' ]
* 2000x blockchain.scripthash.subscribe localhost(39.844 s)
* 2000x blockchain.scripthash.subscribe localhost(40.851 s)
* 2000x blockchain.scripthash.subscribe localhost(41.886 s)
* 2000x blockchain.scripthash.subscribe localhost(41.982 s)
* 2000x blockchain.scripthash.subscribe localhost(41.753 s)
* 2000x blockchain.scripthash.subscribe localhost(41.950 s)
* 2000x blockchain.scripthash.subscribe localhost(42.162 s)
* 2000x blockchain.scripthash.subscribe localhost(41.867 s)
* 2000x blockchain.scripthash.subscribe localhost(42.210 s)
* 2000x blockchain.scripthash.subscribe localhost(42.504 s)
```

***This PR - DEBUG***
```
Connecting to  [ 60001, 'localhost', 'tcp' ]
* 2000x blockchain.scripthash.subscribe localhost(35.505 s)
* 2000x blockchain.scripthash.subscribe localhost(16.546 s)
* 2000x blockchain.scripthash.subscribe localhost(16.571 s)
* 2000x blockchain.scripthash.subscribe localhost(16.538 s)
* 2000x blockchain.scripthash.subscribe localhost(16.570 s)
* 2000x blockchain.scripthash.subscribe localhost(16.523 s)
* 2000x blockchain.scripthash.subscribe localhost(16.529 s)
* 2000x blockchain.scripthash.subscribe localhost(16.523 s)
* 2000x blockchain.scripthash.subscribe localhost(16.595 s)
* 2000x blockchain.scripthash.subscribe localhost(16.535 s)
```

***Before this PR - RELEASE***
```
Connecting to  [ 60001, 'localhost', 'tcp' ]
* 2000x blockchain.scripthash.subscribe localhost(23.008 s)
* 2000x blockchain.scripthash.subscribe localhost(4.692 s)
* 2000x blockchain.scripthash.subscribe localhost(4.708 s)
* 2000x blockchain.scripthash.subscribe localhost(4.653 s)
* 2000x blockchain.scripthash.subscribe localhost(4.665 s)
* 2000x blockchain.scripthash.subscribe localhost(4.689 s)
* 2000x blockchain.scripthash.subscribe localhost(4.713 s)
* 2000x blockchain.scripthash.subscribe localhost(4.689 s)
* 2000x blockchain.scripthash.subscribe localhost(4.680 s)
* 2000x blockchain.scripthash.subscribe localhost(4.674 s)
```

***This PR - RELEASE***
```
Connecting to  [ 60001, 'localhost', 'tcp' ]
* 2000x blockchain.scripthash.subscribe localhost(21.723 s)
* 2000x blockchain.scripthash.subscribe localhost(3.604 s)
* 2000x blockchain.scripthash.subscribe localhost(3.593 s)
* 2000x blockchain.scripthash.subscribe localhost(3.618 s)
* 2000x blockchain.scripthash.subscribe localhost(3.614 s)
* 2000x blockchain.scripthash.subscribe localhost(3.613 s)
* 2000x blockchain.scripthash.subscribe localhost(3.608 s)
* 2000x blockchain.scripthash.subscribe localhost(3.611 s)
* 2000x blockchain.scripthash.subscribe localhost(3.621 s)
* 2000x blockchain.scripthash.subscribe localhost(3.616 s)
```

To test correctness I have written [this integration test](https://github.com/BitcoinUnlimited/BitcoinUnlimited/commit/73c5e4be01e0ae39080e3fcb1cf6f17fbd9c2583), but to port it to this repository https://github.com/romanz/electrs/issues/168 needs to be solved.

### DIFF
--- a/src/bin/electrs.rs
+++ b/src/bin/electrs.rs
@@ -62,12 +62,28 @@ fn run_server(config: &Config) -> Result<()> {
     let query = Query::new(app.clone(), &metrics, tx_cache, config.txid_limit);
 
     let mut server = None; // Electrum RPC server
+
+    // If we see this more new blocks than this, don't look for scripthash
+    // changes. Just have clients reconnect.
+    //
+    // This many header changes means we're either not fully synced, or there
+    // has been abnormally large reorg of the blockchain.
+    let MAX_SCRIPTHASH_BLOCKS = 10;
+
     loop {
-        app.update(&signal)?;
-        query.update_mempool()?;
-        server
-            .get_or_insert_with(|| RPC::start(config.electrum_rpc_addr, query.clone(), &metrics))
-            .notify(); // update subscribed clients
+        let (headers_changed, new_tip) = app.update(&signal)?;
+        let txs_changed = query.update_mempool()?;
+        let rpc = server
+            .get_or_insert_with(|| RPC::start(config.electrum_rpc_addr, query.clone(), &metrics));
+        if headers_changed.len() > MAX_SCRIPTHASH_BLOCKS {
+            rpc.disconnect_clients();
+        }
+        else {
+            rpc.notify_scripthash_subscriptions(&headers_changed, txs_changed);
+        }
+        if let Some(header) = new_tip {
+            rpc.notify_subscriptions_chaintip(header);
+        }
         if let Err(err) = signal.wait(Duration::from_secs(5)) {
             info!("stopping server: {}", err);
             break;
--- a/src/rpc.rs
+++ b/src/rpc.rs
@@ -2,10 +2,11 @@ use bitcoin::blockdata::transaction::Transaction;
 use bitcoin::consensus::encode::{deserialize, serialize};
 use bitcoin_hashes::hex::{FromHex, ToHex};
 use bitcoin_hashes::sha256d::Hash as Sha256dHash;
+use bitcoin_hashes::Hash;
 use error_chain::ChainedError;
 use hex;
 use serde_json::{from_str, Value};
-use std::collections::HashMap;
+use std::collections::{HashMap, HashSet};
 use std::io::{BufRead, BufReader, Write};
 use std::net::{Shutdown, SocketAddr, TcpListener, TcpStream};
 use std::sync::mpsc::{Sender, SyncSender, TrySendError};
@@ -13,8 +14,10 @@ use std::sync::{Arc, Mutex};
 use std::thread;
 
 use crate::errors::*;
+use crate::index::compute_script_hash;
 use crate::metrics::{Gauge, HistogramOpts, HistogramVec, MetricOpts, Metrics};
 use crate::query::{Query, Status};
+use crate::util::FullHash;
 use crate::util::{spawn_thread, Channel, HeaderEntry, SyncChannel};
 
 const ELECTRS_VERSION: &str = env!("CARGO_PKG_VERSION");
@@ -203,6 +206,9 @@ impl Connection {
         let status = self.query.status(&script_hash[..])?;
         let result = status.hash().map_or(Value::Null, |h| json!(hex::encode(h)));
         self.status_hashes.insert(script_hash, result.clone());
+        self.stats
+            .subscriptions
+            .set(self.status_hashes.len() as i64);
         Ok(result)
     }
 
@@ -237,10 +243,6 @@ impl Connection {
         let tx = hex::decode(&tx).chain_err(|| "non-hex tx")?;
         let tx: Transaction = deserialize(&tx).chain_err(|| "failed to parse tx")?;
         let txid = self.query.broadcast(&tx)?;
-        self.query.update_mempool()?;
-        if let Err(e) = self.chan.sender().try_send(Message::PeriodicUpdate) {
-            warn!("failed to issue PeriodicUpdate after broadcast: {}", e);
-        }
         Ok(json!(txid.to_hex()))
     }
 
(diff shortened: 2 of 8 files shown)


src/rpc.rs
```diff
@@ -573,8 +606,105 @@ impl RPC {
...
+
+        let txn = self.query.load_txn(txid, blockhash)?;
+
+        let scripthashes: Vec<FullHash> = txn
+            .output
+            .iter()
+            .map(|o| compute_script_hash(&o.script_pubkey[..]))
+            .collect();
+
+        for s in scripthashes {
```
oleiba: For a new transaction, you are calling this function with its _spent_ transaction id.
So I believe you unintentionally send notifications for sibling outputs of the _spent_ transaction (instead of focusing only on the relevant output which was spent in the new transaction).

src/rpc.rs
```diff
@@ -573,8 +606,105 @@ impl RPC {
...
+        }
+
+        if notify_inputs {
+            for txin in txn.input {
+                if txin.previous_output.is_null() {
+                    continue;
+                }
+                let id: &Sha256dHash = &txin.previous_output.txid;
+
+                if let Err(e) = self.try_notify_subscriptions_for_tx(id, None, notified, false) {
```
oleiba: You are calling `try_notify_subscriptions_for_tx` for a new transaction's spending transactions.
Recall that you are parsing the script-hashes of all the outputs in this function (line 633).
So I believe you unintentionally send notifications for sibling outputs of the spent transaction (instead of focusing only on the relevant output which was spent in the new transaction).

src/bin/electrs.rs
```diff
@@ -62,12 +62,28 @@ fn run_server(config: &Config) -> Result<()> {
...
-        app.update(&signal)?;
-        query.update_mempool()?;
-        server
-            .get_or_insert_with(|| RPC::start(config.electrum_rpc_addr, query.clone(), &metrics))
-            .notify(); // update subscribed clients
+        let (headers_changed, new_tip) = app.update(&signal)?;
+        let txs_changed = query.update_mempool()?;
+        let rpc = server
+            .get_or_insert_with(|| RPC::start(config.electrum_rpc_addr, query.clone(), &metrics));
+        if headers_changed.len() > MAX_SCRIPTHASH_BLOCKS {
```
oleiba: Is the server being open before Electrum reaches the tip of its daemon?
Id so, then isn't this check redundant? (clients will be disconnected anyway. And if there's a re-org, why not sending a notification?)

src/rpc.rs
```diff
@@ -573,8 +606,105 @@ impl RPC {
         }
     }
 
-    pub fn notify(&self) {
-        self.notification.send(Notification::Periodic).unwrap();
+    /// Attempt to notify scripthash subscriptions affected by transaction.
+    /// The `notified` hashset is an optimization to avoid double notifications.
```
oleiba: Do you think it would be easier if instead of passing an accumulated `HashSet` of notified script hashes, this function would return a set of script hashes to the caller, and the _caller_ would be the one iterating over it and sending the notifications?

src/rpc.rs
```diff
@@ -573,8 +606,105 @@ impl RPC {
...
+
+        for s in scripthashes {
+            if let Err(e) = self.notification.send(Notification::ScriptHashChange(s)) {
+                trace!("ScriptHash change notification failed {}", e);
+            }
+        }
+
+        if notify_inputs {
+            for txin in txn.input {
+                if txin.previous_output.is_null() {
```
oleiba: Can we use `is_coin_base()` here, or are there more cases?

src/rpc.rs
```diff
@@ -237,10 +243,6 @@ impl Connection {
         let tx = hex::decode(&tx).chain_err(|| "non-hex tx")?;
         let tx: Transaction = deserialize(&tx).chain_err(|| "failed to parse tx")?;
         let txid = self.query.broadcast(&tx)?;
-        self.query.update_mempool()?;
```
oleiba: Doesn't removing it mean that you will not get an immediate notification, but wait for the periodic 5 seconds?
Maybe we can get the changed txids in `update_mempool`, and then call `on_scripthash_change` after parsing the transactions?...

src/rpc.rs
dagurval: Good observation

src/rpc.rs
dagurval: Maybe? :-)

src/rpc.rs
dagurval: Yes, but with PeriodicUpdate taking 3+ seconds to begin with, there isn't much difference. Also, if within the scope of this project, I want to add optional support for ZMQ, which would remove the 5 second delay.

src/bin/electrs.rs
dagurval: > Is the server being open before Electrum reaches the tip of its daemon?

Yes, with my testing clients can connect before we're synced up. It takes a very long time to sync the final blocks if we produce notifications at the same time (in the normal case it doesn't take long, because we've already seen and cached the transactions when they entered mempool).

Also there is also no way to know from here if any clients are connected, so this was the simplest solution. 

As for reorg, we can pretty much assume that a >= 10 block won't happen on mainnet, and if we disconnect the clients on such a big reorg its no big deal. We can also increase `MAX_SCRIPTHASH_BLOCKS` if its too low.

dagurval:
Thanks for the review @oleiba!

dagurval:
I see that CI is failing on syntax errors, but I've noticed that @romanz tends to rewrite everything anyway, so its better to not force push changes too much :-)

src/rpc.rs
oleiba: from UX perspective - getting the sender's tx indexed is important for fetching history and unspent immediately after the transfer.
I think there's a non-negligible difference between 3+ seconds (I think it's less actually when you force a mempool-only update not within the 5 seconds interval) and waiting this _plus_ waiting for the interval to occur (8 seconds?)...

src/rpc.rs
oleiba: Ok, I get that :)

dagurval:
This was added with fix for problem pointed out by oleiba & removal of recursion [downstream](https://github.com/BitcoinUnlimited/ElectrsCash/pull/31). 

This is probably too invasive. Maybe go for #186 here instead.

Issue #146, opened 2019-05-30
Implement an efficient subscription tracker


romanz:
Current implementation has bad performance on embedded devices when tracking large amount of addresses (e.g. #143).

santzi:
I noticed also this bad performance if address amount >500 (>2000 is terrible).
CPU FX-8350 / 16GB RAM / SSD
CPU load is 100% long time for this electrs process.
I think this #73 also might increase performance a little bit
At the moment there are hundreds RPC Requests on log for single address.

oleiba:
I would like to suggest a short analysis: I believe the issue is due to the heavy DB scan and filtering which happens when which addresses are subscribed (via [`blockchain_scripthash_subscribe`](https://github.com/romanz/electrs/blob/9df66d97e3d3c3f12ad73d485b768bbaabba3f91/src/rpc.rs#L201)) and later when notifications are updated periodically (via [`update_notifications`](https://github.com/romanz/electrs/blob/9df66d97e3d3c3f12ad73d485b768bbaabba3f91/src/rpc.rs#L335)).

The implementation seems to scan the DB for both all blocks and the mempool, and then filter by address.
This is done for every single subscribed address separately.
I would suggest to refactor it so the process would go over the history (confirmed and mempool) in _one_ pass, and filter transactions by _any_ of the subscribed addresses.

This can also be further optimized, since there's no need to go over blocks prior to the highest block which is common (i.e. same hash) between the current periodic update and the previous (at least if no new addresses has been subscribed since). However, I would say the previous fix alone should already eliminate much of the load.

romanz:
Could you please share the benchmarking results?
I prefer to make sure that we understand the exact performance bottleneck, before starting the optimization.

romanz:
We can continue the discussion on Gitter if you prefer:
https://gitter.im/romanz/electrs?at=5d9edd58464b432fc1f0171e

oleiba:
I currently do not have benchmarking results. This is was an educated assumption based on reviewing the code triggered for every periodic update of notifications.

shesek:
Ugh, sorry, the commit mention was by mistake, must've pasted the wrong url somehow...

gustavonalle:
In my quad core box, 1 core goes exclusively for electrs, 100% all the time. The wallet has hundreds of addresses. Any workaround for this until the new implementation is not ready?

gustavonalle:
Profile info when watching hundreds of addresses. Mostly stuck on RocksDB  

```
[...]

PR #186, opened 2019-10-31
Scale script hashes subscriptions

oleiba:
Relates to issue #146.

Main idea - 
1) Only upon a new batch of mempool transactions or a new block, the mempool tracker and blockchain poller (respectively) will parse the participating addresses in the new transactions.
2) The list of participating addresses will be passed to each Connection.
3) Each Connection will filter and query the status only of its subscribed addresses which participated in this recent batch of transactions.

What was before:
Every n seconds each connection would query the status of **all** its subscribed addresses.

This PR would limit the number of operations in every periodic lookup to be in the order of __new transactions in this period of time__, instead of the order of __subscribed addresses__.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,5 @@
 language: rust
 
-arch:
-  - amd64
-  - arm64
-
 rust:
   - stable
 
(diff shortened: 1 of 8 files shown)


oleiba:
Closing in favor of #185.

PR #190, opened 2019-11-09
Allow setting cookie via CLI flag or ENV var

### LINKED ISSUE #189: Allow cookie to be set via CLI flag or ENV var
>In addition, config files support cookie option to specify cookie - this is not available using command line or environment variables for security reasons (other applications could read it otherwise).

This good practise in general, however there are some scenarios where it may be useful to allow setting cookie via CLI flag or ENV var where it's not a security concern.

For example, in an isolated Docker network I just use the credentials `rpcuser:rpcpassword`. It's easy set between different containers and the RPC port isn't even accessible by the host, it's only internal to the Docker network.

To get this working with electrs I need to create a config file purely to set the cookie which is awkward. It would be great if I could just add `ELECTRS_COOKIE=rpcuser:rpcpassword` to my `docker-compose.yml`.

Are you open to allowing this?

lukechilds:
Resolves #189

### DIFF
--- a/config_spec.toml
+++ b/config_spec.toml
@@ -33,9 +33,6 @@ default = "crate::config::default_daemon_dir()"
 name = "cookie"
 type = "String"
 doc = "JSONRPC authentication cookie ('USER:PASSWORD', default: read from ~/.bitcoin/.cookie)"
-# Force the user to use config file in order to avoid password leaks
-argument = false
-env_var = false
 
 [[param]]
 name = "network"


oleiba:
I believe it is intentionally not available as an ENV variable or CLI flag.
from [Usage.md](https://github.com/romanz/electrs/blob/master/doc/usage.md):
```
In addition, config files support cookie option to specify cookie - this is not available using command line or environment variables for security reasonns (other applications could read it otherwise).
```

lukechilds:
Yeah I'm aware it's intentional, see related issue for context: https://github.com/romanz/electrs/issues/189

romanz:
Please try to use `--cookie-file` (see 5549287).

romanz:
Ping :)

lukechilds:
@romanz sorry, missed this!

I was just benchmarking electrs/ElectrumX against each other in a virtual machine for research purposes.

I've finished my tests and destroyed the machine now so this isn't something I need right now or can easily test.

`--cookie-file` definitely looks like a good solution to the problem though!

PR #191, opened 2019-11-10
Implemented custom cookie file (untested!)

### LINKED ISSUE #176: Custom cookie file path
It'd be nice to allow for custom cookie file path if the user uses non-default one. I will make a PR if nobody is faster than myself. :) ( Ihave some other important things to do right now.)

### LINKED ISSUE #189: Allow cookie to be set via CLI flag or ENV var

Kixunil:
This change allows the user to specify a custom cookie file, which is then
used instead of `~/.bitcoin/.cookie`. This resolves situations when the
user wants to have the cookie file in non-standard path.

Aside from that, the code now pre-computes the default path, improving
the performance by avoiding allocation (and copying). Unfortunately, due
to limitations of Rust, the code doesn't print out cookie configuration
anymore. This however might be safer, since the cookie isn't printed,
and thus doesn't end up in some readable logs by accident.

Closes #176
Closes #189

### DIFF
--- a/config_spec.toml
+++ b/config_spec.toml
@@ -32,10 +32,16 @@ default = "crate::config::default_daemon_dir()"
 [[param]]
 name = "cookie"
 type = "String"
-doc = "JSONRPC authentication cookie ('USER:PASSWORD', default: read from ~/.bitcoin/.cookie)"
+doc = "JSONRPC authentication cookie ('USER:PASSWORD', default: read from cookie file)"
 # Force the user to use config file in order to avoid password leaks
 argument = false
 env_var = false
+
+[[param]]
+name = "cookie_file"
+type = "std::path::PathBuf"
+doc = "JSONRPC authentication cookie file (default: ~/.bitcoin/.cookie)"
+# This is safe to configure on command line.
 
 [[param]]
 name = "network"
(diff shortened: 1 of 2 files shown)


Kixunil:
I believe this should work, but didn't test it since I don't have any suitable testing environment at hand right now.

@lukechilds could you help with testing?

Kixunil:
Looks like there's CI bug.

romanz:
Thanks a lot, tested and merged via https://github.com/romanz/electrs/commit/5549287ac6389d02c6410714d5864935b6d2c92e.

PR #195, opened 2019-11-15
Clean up RPC handles when connections are closed

champo:
Fixes https://github.com/romanz/electrs/issues/188

We have a setup with 2 electrs running behind a TCP load balancer. The LB do a regular health check to electrs RPC port by seeing if they can open a socket, as soon as it's established they have no need for it and so they close it.

The RPC code was holding onto `JoinHandle` and `SyncSender` for each connection ever created. That slowly added up to increased memory usage and file descriptors.

This fix was tested by monitoring electrs memory usage while opening and closing RPC sockets in a loop for about 10 minutes.

PR #196, opened 2019-11-15
Report cache usage metrics to prometheus

champo:
During development of #195 it was a bit confusing that memory usage reported to prometheus sometimes jumped. This was most likely due to new transactions being stored in cache. Adding this metric helps understand how memory is being used by electrs.


romanz:
Thanks!

PR #204, opened 2019-12-15
Correct --network arg values in the doc

x1ddos:
After a recent switch to configure_me in ec049b9a, the 'mainnet'
arg value became 'bitcoin'. 2c50791d then updated db dir location
but the doc for the --network arg remained stale.

An alternative is to accept both 'mainnet' and 'bitcoin' but
it would be confusing IMHO.


romanz:
Thanks!

PR #232, opened 2020-04-12
split_off channged to retain

Kixunil:
This changes `split_off`, which does one allocation and copies `n` items
in the best case, `2 * n` items in the worst case to `retain`, which does
zero allocations and copies zero items in the best case and `3 * (n - 1)`
items in worst case (because swap is 3 copies and `retain` has to swap
due to possibility of panic).

I believe this should be net-benefit due to usually small `n`.

**Disclaimer**: didn't have time to test, but it looks obvious enough. :)

### DIFF
--- a/src/rpc.rs
+++ b/src/rpc.rs
@@ -497,14 +497,15 @@ impl RPC {
                 let mut senders = senders.lock().unwrap();
                 match msg {
                     Notification::Periodic => {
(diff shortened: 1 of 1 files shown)


romanz:
Many thanks!
Added a small comment, and rebased: https://github.com/romanz/electrs/commit/669a4b829dff2c5a4affc23f2dad2c0d641c910c.

PR #233, opened 2020-04-12
Actually wait for threads in RPC.

Kixunil:
**Important: untested suggestion of how to solve the problem!**

This change implements non-blocking cleanup of threads. It achieves it
by cleaning up periodically based on thread ID, that gets sent from dead
thread over a channel. Aside from internal locks in the channel, there
are no other locks. The cleanup also happens only after a new connection
is accepted, but hopefully won't be an issue, unless there are many
connections that die and then nothing connects for a long time. A final
cleanup happens when the thread is finishing.


romanz:
Many thanks!

PR #234, opened 2020-04-13
Base Docker image on debian:buster-slim

### LINKED ISSUE #226: electrs exits with code 135
I have the following Docker Compose setup:

```yaml
version: "2"
services:

  bitcoind:
    image: lukechilds/bitcoind:v0.19.1
    stop_grace_period: 10m
    restart: always
    expose:
      - "8332"
    volumes:
      - ./data/bitcoind:/data/.bitcoin
    command: >-
      -rpcbind=0.0.0.0
      -rpcallowip=0.0.0.0/0
      -disablewallet=1

  electrs:
    image: electrs-app:latest
    stop_grace_period: 10m
    restart: always
    user: root
    ports:
      - "50001:50001"
    volumes:
      - ./data/bitcoind:/root/.bitcoin:ro
      - ./data/electrs:/root
    command: >-
      electrs
      -vvvv
      --timestamp
      --db-dir /root/db
      --daemon-rpc-addr bitcoind:8332
```

Changed to run as root user because `lukechilds/bitcoind` currently runs as root so `electrs-app` needs to be root to read `.bitcoin/.cookie`. I know I should fix that but it's fine for just playing around.

Anyway, I don't think it should be root that's causing issues, I'm sure (I think?) this exact config has worked before, but for some reason when I try to run it now I get:

```
pi@pinode:~ $ docker-compose up electrs
Starting pi_electrs_1 ... done
Attaching to pi_electrs_1
electrs_1   | Config { log: StdErrLog { verbosity: Trace, quiet: false, timestamp: Millisecond, modules: [], writer: "stderr", color_choice: Auto }, network_type: bitcoin, db_path: "/root/db/mainnet", daemon_dir: "/root/.bitcoin", daemon_rpc_addr: V4(172.18.0.2:8332), electrum_rpc_addr: V4(127.0.0.1:50001), monitoring_addr: V4(127.0.0.1:4224), jsonrpc_import: false, index_batch_size: 100, bulk_index_threads: 4, tx_cache_size: 10485760, txid_limit: 100, server_banner: "Welcome to electrs 0.8.3 (Electrum Rust Server)!", blocktxids_cache_size: 10485760 }
electrs_1   | 2020-03-30T18:49:48.987+00:00 - DEBUG - Server listening on 127.0.0.1:4224
electrs_1   | 2020-03-30T18:49:48.989+00:00 - DEBUG - Running accept thread
electrs_1   | 2020-03-30T18:49:48.996+00:00 - INFO - NetworkInfo { version: 190100, subversion: "/Satoshi:0.19.1/", relayfee: 0.00001 }
electrs_1   | 2020-03-30T18:49:48.997+00:00 - INFO - BlockchainInfo { chain: "main", blocks: 623645, headers: 623645, bestblockhash: "0000000000000000000f359f884320a688b7c9d29b2aa3d77e25b40d0c2d149a", pruned: false, initialblockdownload:
false }
electrs_1   | 2020-03-30T18:49:48.998+00:00 - DEBUG - opening DB at "/root/db/mainnet"
pi_electrs_1 exited with code 135
```

Is there any way to get any more debug to see why it's failing? Does code 135 mean something specific?

Thanks!

lukechilds:
Resolves #226.

The current Docker image has a strange bug when being built on 32 bit ARM devices causing RocksDB to be built incorrectly and `electrs` to exit as soon as it's started.

Updating to Debian Buster seems to fix the bug somewhere in the C++ toolchain. There is no official Buster Docker image for Rust 1.34.0 so instead I've used Debian Buster as the base image and manually installed Rust 1.34.0.

I've also added `RUST_VERSION` as a build arg which is passed through to `rustup`.

This means the image easily be changed to support any Rust version supported by `rustup` at build time by passing in a build arg or by changing the default in the Dockerfile.

e.g:
```
# Builds with Rust 1.34.0 by default
$ docker build -t electrs .

# Manually specify a Rust version
$ docker build -t electrs --build-arg RUST_VERSION=1.40.0 .

# Use the latest stable Rust version
$ docker build -t electrs --build-arg RUST_VERSION=stable .
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,10 +1,24 @@
-FROM rust:1.34.0-slim
+FROM debian:buster-slim
 
 RUN apt-get update
 RUN apt-get install -y clang cmake
 RUN apt-get install -y libsnappy-dev
+RUN apt-get install -y curl
 
 RUN adduser --disabled-login --system --shell /bin/false --uid 1000 user
+
+ARG RUST_VERSION=1.34.0
+ENV RUSTUP_HOME /usr/local/rustup
+ENV CARGO_HOME /usr/local/cargo
+ENV PATH $CARGO_HOME/bin:$PATH
+
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- \
+  -y \
+  --verbose \
+  --profile minimal \
+  --default-toolchain $RUST_VERSION
+
+RUN chmod -R a+w $RUSTUP_HOME $CARGO_HOME
 
 USER user
 WORKDIR /home/user


Dockerfile
```diff
@@ -1,11 +1,25 @@
...
+ARG RUST_VERSION=1.34.0
+ENV RUSTUP_HOME /usr/local/rustup
+ENV CARGO_HOME /usr/local/cargo
+ENV PATH $CARGO_HOME/bin:$PATH
+
+RUN curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh -s -- \
+  -y \
+  --verbose \
+  --profile minimal \
+  --default-toolchain $RUST_VERSION
```
Kixunil, line 19: I meant to install it using `apt-get` to get signature validation for free. If you prefer rustup, then I'd suggest doing what the official Docker image does and check the hashes.

Dockerfile
```diff
@@ -1,11 +1,25 @@
...
+
+RUN chmod -R a+w $RUSTUP_HOME $CARGO_HOME
```
Kixunil, line 21: This looks strange, is it a standard thing to do in docker?

lukechilds: No, it's not a particularly standard thing to do in Docker, normally you would compile a binary then copy it to `usr/local/bin` or similar.

Not very familiar with Cargo but it seems to want it's own build dir with other data that you then add to your path.

Without changing the permissions the `user` account doesn't have permission to write to `$CARGO_HOME` during installation.

The official Rust images do it like this too: https://github.com/rust-lang/docker-rust/blob/8bab191937fcf23569d3a3c31103c1c6f7f2947e/1.42.0/buster/slim/Dockerfile#L30

I guess we could just build as root in `/root/.cargo/bin` and then copy the `electrs` binary to `/usr/local/bin` which should allow `user` to execute it.

Dockerfile
lukechilds: Good point regarding `apt` over `rustup`.

However regarding rustup, it doesn't seem possible to get a secure Rust environment via `rustup` which was why I intentionally didn't bother with hashes.

For example the official Docker image only checks the hash of `rustup-init.sh`. This is pointless because:
1. `rustup-init.sh` then downloads `rustup` but doesn't check hashes/sigs.
2. `rustup` doesn't check signatures when installing Rust versions.

Se verifying `rustup-init.sh` doesn't guarantee anything about `rustup`, and even if you get an honest `rustup`, that doesn't even guarantee you get an honest `rustc`/`cargo`!

But you're right about preferring `apt`, that would definitely be better.

lukechilds:
@Kixunil Good points in your feedback. I just went with `rustup` due to the fact that it allows us to change the Rust version easily and that's also what the official Docker Rust images are using.

But I'm also not a fan of no signature verification and clunky `$CARGO_HOME` workarounds.

Installing `cargo` from `apt` and moving binaries into `/usr/local/bin` and not having to worry about `$CARGO_HOME` permissions seems like an improvement over the official Image.

I'll wait to hear back from @romanz first re https://github.com/romanz/electrs/issues/226#issuecomment-612776828. Maybe a super secure Docker env is outside the scope of this project and just a simple functional Dockerfile using the official Rust images is adequate which we now have in https://github.com/romanz/electrs/commit/7d23da5ffcc2e0e6a65eef6354a665287d9741fc.

Kixunil:
Right. If secure docker is out of scope, this should be documented.

romanz:
Thanks!
Merged using #235 instead.

PR #238, opened 2020-04-19
Migration to new bitcoin hash type system; bumping to the latest rust-bitcoin release

dr-orlovsky:
After my work on new bitcoin crate hash type system last year <https://github.com/rust-bitcoin/rust-bitcoin/pull/349> I find many project depending on the upstream `bitcoin` crate being unable to update to the latest version without spending a lot of effort on type replacements (plus some additional code editing in cases like merklization). Due to my work on adding support to Signet (which was implemented in rust-bitcoin fork containing the new type system) I had to do this work, so I propose it for your consideration as a PR here. Pls review it before significant changes to the master, since it will take many hours of my time again to do any rebase, which I do not have.

New hash type system from bitcoin crate does not cover all cases for merkle values; and the same function may be applied to different hash types. Thus, I have used generic types to abstract the logic.

`Txid` merklization operates with TxMerkleNode type; `BlockHash` merklization does not introduce a new type: the same design decision was made in the original work on new bitcoin crate type system since it is used in a single case, and there is no point in introducing a special designated hash type.

"Script hashes" met through the code are in fact has nothing to do with P2SH script hashes and `ScripHash` type; but instead are custom pubkeyScript hashes used in RPC queries. This data structures are left of Sha256dHash type since there is no corresponding type defined in bitcoin crate; this type is specific to Electrum X protocol. Corresponding new type can be implemented in the project later with `hash_newtype!` macro in the same way it was done in bitcoin crate.

### DIFF
--- a/src/query.rs
+++ b/src/query.rs
@@ -1,7 +1,7 @@
 use bitcoin::blockdata::transaction::Transaction;
 use bitcoin::consensus::encode::deserialize;
+use bitcoin::hash_types::{Txid, BlockHash, TxMerkleNode};
 use bitcoin_hashes::hex::ToHex;
-use bitcoin_hashes::sha256d::Hash as Sha256dHash;
 use bitcoin_hashes::Hash;
 use crypto::digest::Digest;
 use crypto::sha2::Sha256;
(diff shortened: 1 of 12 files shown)


src/query.rs
```diff
@@ -428,7 +429,11 @@ impl Query {
             .iter()
             .position(|txid| txid == tx_hash)
             .chain_err(|| format!("missing txid {}", tx_hash))?;
-        let (branch, _root) = create_merkle_branch_and_root(txids, pos);
+        let tx_nodes: Vec<TxMerkleNode> = txids
+            .into_iter()
+            .map(|txid| TxMerkleNode::from_inner(txid.into_inner()))
+            .collect();
```
Kixunil, line 435: I don't like this allocation, but removing it is not trivial. I might look at it later and see if I can optimize it.

dr-orlovsky: Well, I think it will be nice to have direction conversion functions for such cases which are optimized by performance – and make them part of the hash type system in `rust-bitcoin`

Kixunil: I was thinking more in terms of replacing vecs with iterators, but not really sure how applicable it is in this case. I managed to refactor those in the past, so I might try again. :)

dr-orlovsky: It seems that have to be a PR to rust-bitcoin though...

romanz:
Many thanks for the contribution, much appreciated!

Issue #259, opened 2020-06-10
question about store.rs


kenshin-samourai:
hey

i apologize in advance for this stupid question but i don't understand the purpose of this line
https://github.com/romanz/electrs/blob/master/src/store.rs#L77

do we try to set self.opts.bulk_import?

romanz:
Thanks for the question!
Actually, it's indeed not clear enough from the code - so I'll try to explain:

`bulk_import` is a flag for choosing the initial block import mode.
If it's `true`, we read the `blk*.dat` files and write them into the DB (with some options being different for performance reasons - e.g. no flushing to disk and disabled compaction).
After we finish the initial block import, we flip it to `false`, flush the writes and compact the DB:
https://github.com/romanz/electrs/blob/1ca25531b7d74d7bcd7f5cbd9d0bcaf763474220/src/store.rs#L184-L188

This mode is used at the first time `electrs` runs, so in the next runs it should use regular block import (via bitcoind JSONRPC interface).

Hope this helps!

kenshin-samourai:
thanks for this explanation @romanz. the rationale makes sense.

if my understanding is correct:
- after the initial block import, writes should be flushed to disk automatically (https://github.com/romanz/electrs/blob/master/src/store.rs#L157)
- when electrs is launched, the dbstore object is always initialized with bulk_import set to true (https://github.com/romanz/electrs/blob/master/src/store.rs#L69).
  this value should be set to false when enable_compaction() is called.

i'm not not a rust expert (far from that actually) but my understanding is that line 77 sets bulk_import to false on a clone of self.opts and not on self.opts itself. thus, flushes to disk are never activated in write(). do i miss something about cloned objects in rust?

i've made some tests on an instance with jsonrpc_import set to true and the compaction flag stored in database, At each startup, electrs processes again all the blocks created after the compaction flag was stored in database, as if they were never written on disk.

romanz:
Many thanks for finding this!
I think it is the cause for https://github.com/romanz/electrs/issues/249.

kenshin-samourai:
good catch! thanks for you work. electrs is a great project!

PR #260, opened 2020-06-11
Fix missing flush after enabling compaction

romanz:
It should be `self.opts.bulk_import = false`.
Previous code did that on a clone:

    let mut opts = self.opts.clone();
    opts.bulk_import = false;

Found at #259 by @kenshin-samourai - thanks!

### DIFF
--- a/src/bin/electrs.rs
+++ b/src/bin/electrs.rs
@@ -40,22 +40,26 @@ fn run_server(config: &Config) -> Result<()> {
         &metrics,
     )?;
     // Perform initial indexing from local blk*.dat block files.
-    let store = DBStore::open(&config.db_path, /*low_memory=*/ config.jsonrpc_import);
+    let mut store = DBStore::open(&config.db_path, /*low_memory=*/ config.jsonrpc_import);
     let index = Index::load(&store, &daemon, &metrics, config.index_batch_size)?;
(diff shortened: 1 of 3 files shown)


romanz:
CC: @DrFredPhD @pd1983

DrFredPhD:
I'm going to try building this branch now and test if the db write issue is fixed.

DrFredPhD:
Unfortunately does not seem to fix issue. Built this branch and fully reindexed database. After indexing and initial compaction was complete successfully connected my electrum wallet. Restarted the server and electrs fails to exit gracefully and begins full compaction again.

romanz:
Very strange... 
I've tried to reproduce this issue on my x86 machine but it seems to work as intended.

BTW, could you try to reproduce it with `regtest`?
```bash
$ ./bitcoind -regtest &
$ ./bitcoin-cli -regtest generatetoaddress 100 mkHS9ne12qx9pS9VojpwU5xtRd4T7X7ZUt   # create 100 blocks, sending reward to a dummy address
$ cargo run --release -- --jsonrpc-import --network=regtest
```

romanz:
@DrFredPhD 
Could you please test the new electrs WIP version?

It can be built using the instructions here:
https://github.com/romanz/electrs/issues/308

I hope it should resolve this issue :)

Issue #261, opened 2020-06-17
build fails on debian 10.4 amd64


oitdmser:
looking to build Electrs as per your [guide.] (https://github.com/romanz/electrs/blob/master/doc/usage.md) Ran the following:

```
apt update
apt install cargo clang cmake
```

cloned the repo, but when I run `cargo build --release` I get a string of errors as follows:

```
error[E0658]: use of unstable library feature 'alloc': this library is unlikely to be stabilized in its current form or name (see issue #27783)
error[E0599]: no function or associated item named `uninit` found for type `core::mem::MaybeUninit<_>` in the current scope
error[E0658]: use of unstable library feature 'maybe_uninit' (see issue #53491)
error: aborting due to 9 previous errors

Some errors occurred: E0599, E0658.

rustic --version
rustc 1.34.2
cargo --version
cargo 1.34.0
```

Kixunil:
Do you have to use the version from master instead of the latest official release? If not, try `git checkout v0.8.3` and build it then.

@romanz  in 687ae6c65b8e3c3c7856c1845d6d4634ed58805f `error-chain` and `backtrace` were updated, which brought in dependency on `addr2line` which in turn added dependency on `gimli`, which uses `MaybeUninit`, breaking compatibility with Rust 1.34. Looks like downgrading these things would help?

oitdmser:
I ran 
git checkout -f tags/v0.8.3
cargo build --release

and it's now building successfully, thanks for that tip. It might be helpful for the usage guide to mention something about master branch versus a specific release?

romanz:
Thanks for opening the issue!
Will fix `master` soon :)

romanz:
Should be fixed by https://github.com/romanz/electrs/pull/263.

PR #263, opened 2020-06-18
Fix dependencies for stable Debian

romanz:
Following https://github.com/romanz/electrs/issues/261.


romanz:
Tested locally:
```bash
$ cargo +1.34.0 build --release 
    Finished release [optimized] target(s) in 0.10s
```

Issue #265, opened 2020-06-27
Transaction fee should be reported


romanz:
https://github.com/spesmilo/electrum/issues/6289

romanz:
Should be fixed by #266

PR #266, opened 2020-06-28
Return fee for unconfirmed transactions history

romanz:
Following https://electrumx-spesmilo.readthedocs.io/en/latest/protocol-methods.html#blockchain.scripthash.get_mempool

Otherwise, the client will disconnect:
https://github.com/spesmilo/electrum/issues/6289


romanz:
Fixes #265

PR #269, opened 2020-06-30
Add a `blocks_dir` option analogous to bitcoind's `-blocksdir`

darosior:
I used the quite convenient `-blocksdir` option (https://github.com/bitcoin/bitcoin/pull/12653) on my `bitcoind`, but had to fall back to symlinking the `blocks/` to be able to use Electrs with `blk`s on a different disk.

This makes Electrs aware that the `blk`s might not always be in `<config_dir>/blocks/`.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -128,6 +128,7 @@ pub struct Config {
     pub network_type: Network,
     pub db_path: PathBuf,
     pub daemon_dir: PathBuf,
+    pub blocks_dir: PathBuf,
     pub daemon_rpc_addr: SocketAddr,
     pub electrum_rpc_addr: SocketAddr,
     pub monitoring_addr: SocketAddr,
@@ -150,6 +151,10 @@ fn default_daemon_dir() -> PathBuf {
     });
     home.push(".bitcoin");
     home
+}
+
+fn default_blocks_dir(daemon_dir: &Path) -> PathBuf {
+    daemon_dir.join("blocks")
 }
 
 fn create_cookie_getter(
@@ -230,6 +235,10 @@ impl Config {
             Network::Regtest => config.daemon_dir.push("regtest"),
         }
 
+        let blocks_dir = config.blocks_dir.unwrap_or(
+            default_blocks_dir(&config.daemon_dir)
+        );
+
         let cookie_getter =
             create_cookie_getter(config.cookie, config.cookie_file, &config.daemon_dir);
 
@@ -260,6 +269,7 @@ impl Config {
             network_type: config.network,
             db_path: config.db_dir,
             daemon_dir: config.daemon_dir,
+            blocks_dir,
             daemon_rpc_addr,
             electrum_rpc_addr,
             monitoring_addr,
@@ -302,6 +312,7 @@ debug_struct! { Config,
     network_type,
     db_path,
     daemon_dir,
+    blocks_dir,
     daemon_rpc_addr,
     electrum_rpc_addr,
     monitoring_addr,
--- a/src/daemon.rs
+++ b/src/daemon.rs
@@ -288,6 +288,7 @@ impl Counter {
 
 pub struct Daemon {
     daemon_dir: PathBuf,
+    blocks_dir: PathBuf,
     network: Network,
     conn: Mutex<Connection>,
     message_id: Counter, // for monotonic JSONRPC 'id'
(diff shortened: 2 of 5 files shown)


src/config.rs
```diff
@@ -152,6 +153,12 @@ fn default_daemon_dir() -> PathBuf {
     home
 }
 
+fn default_blocks_dir(daemon_dir: &PathBuf) -> PathBuf {
+    let mut blocks_dir = daemon_dir.clone();
+    blocks_dir.push("blocks");
```
Kixunil, line 158: This is slightly more complicated than it could be. I suggest accepting `&Path` instead of `&PathBuf` and then just using `join()` In other words, this:

```rust
fn default_blocks_dir(daemon_dir: &Path) -> PathBuf {
    daemon_dir.join("blocks")
}
```

src/config.rs
```diff
@@ -230,6 +237,10 @@ impl Config {
             Network::Regtest => config.daemon_dir.push("regtest"),
         }
 
+        let blocks_dir = config.blocks_dir.unwrap_or(
+            default_blocks_dir(&config.daemon_dir)
```
Kixunil, line 241: `unwrap_or_else` would avoid needless allocation in case `blocks_dir` is specified.

src/daemon.rs
```diff
@@ -371,8 +375,7 @@ impl Daemon {
     }
 
     pub fn list_blk_files(&self) -> Result<Vec<PathBuf>> {
-        let mut path = self.daemon_dir.clone();
-        path.push("blocks");
+        let mut path = self.blocks_dir.clone();
         path.push("blk*.dat");
```
Kixunil, line 379: I'd suggest changing these two lines to `let path = self.blocks_dir.join("blk*.dat");` it's more obvious and more efficient.

src/config.rs
darosior: Right, I just moved the code from `list_blk_files()`. But other words is way nicer, thanks !

src/config.rs
darosior: Hmm yep but it takes a closure and would require to move config and borrow it (as we derive the default blocks dir from the daemon dir).

darosior:
Amended as per @Kixunil 's review

romanz:
Many thanks for the contribution and the review!

src/config.rs
Kixunil: Borrowing inside closure should be enough. Am I missing something?

darosior: Hmm I'm still learning Rust so I was interested in trying this out. I couldn't come up with something smarter than another allocation to avoid `move`ing `config`:
```patch
diff --git a/src/config.rs b/src/config.rs
index 1bc4dda..a9975e9 100644
--- a/src/config.rs
+++ b/src/config.rs
@@ -235,8 +235,9 @@ impl Config {
             Network::Regtest => config.daemon_dir.push("regtest"),
         }
 
-        let blocks_dir = config.blocks_dir.unwrap_or(
-            default_blocks_dir(&config.daemon_dir)
+        let daemon_dir = config.daemon_dir.clone();
+        let blocks_dir = config.blocks_dir.unwrap_or_else(||
+            default_blocks_dir(&daemon_dir)
         );
 
         let cookie_getter =
```

Kixunil: I don't see why `let blocks_dir = config.blocks_dir.unwrap_or(|| default_blocks_dir(&config.daemon_dir));` would be problematic, maybe I'll take a look at it some day.

darosior: Because you try to borrow a moved value.

Kixunil: Oh, I see now. It can be solved using additional variable, there's an RFC about making this case work out of the box. I think I will be able to make a PR soon.

Kixunil: Hah, wanted to look at it now, but @romanz managed to do it before me (feel free to laugh imagining my puzzled look :D). If you're still interested in learning, you can take a look at his change: https://github.com/romanz/electrs/commit/521914a853088d5bf4276276df2648d644539f8f#diff-6d8a5e2104954a026de2bfe2b936aea4R239

darosior: Thanks! :)
That's exactly what i came up with above: but in both cases there is a needless an allocation even when `blocks_dir` is specified (the original comment of this thread).

Kixunil: Hmm? What needless allocation? He solved the allocation by changing `ok_or` to `ok_or_else`.

darosior: Oh I read too fast: actually i was wrong he did `let daemon_dir = &config.daemon_dir;` while i came up with `config.daemon_dir.clone();` !

PR #284, opened 2020-07-28
Drop receiver in handle_replies

### LINKED ISSUE #283: Server does not want to shut down when I press Ctrl + C
When I press Ctrl + C to stop the server, it seems to crash, this was not the first time, the last time I closed the terminal abruptly and when I started Electrs again he no longer connected to Electrum or another wallet that supports servers.

I had to delete the DB folder where the indexed files are in order to download everything again which took ~ 5 hours for the server to reconnect the wallets again, what should I do when that happens?

Kixunil:
When `Message::Done` was received before the cleanup loop executed it
led to a deadlock since the loop didn't see the receiving end being
disconnected and it also had strong reference via `Arc` preventing it
from dropping.

This change solves the problem by splitting the channel and only keeping
the sender in `Connection`. The reciver is passed to `handle_replies` as
an extra argument, making sure it will be used exactly once.

The main advantage of this method over trying to keep the receiver
inside `Connection` is that it hanles cases when the thread panicks and
possibly forgetting explicit close in future refactors. Another
advantage is being able to remove one `Arc::clone` and I have an idea
for removing locking on `senders` in the future too.

Closes #283

**Not tested yet**, however, I'm fairly confident in it being OK as it was a simple change and it typechecks. :) I don't think I'll be able to test today.

### DIFF
--- a/src/rpc.rs
+++ b/src/rpc.rs
@@ -7,14 +7,14 @@ use serde_json::{from_str, Value};
 use std::collections::HashMap;
 use std::io::{BufRead, BufReader, Write};
 use std::net::{Shutdown, SocketAddr, TcpListener, TcpStream};
-use std::sync::mpsc::{Sender, SyncSender, TrySendError};
+use std::sync::mpsc::{self, Sender, SyncSender, Receiver, TrySendError};
 use std::sync::{Arc, Mutex};
 use std::thread;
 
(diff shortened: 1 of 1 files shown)


romanz:
Good catch!
Many thanks for the fix :)

PR #288, opened 2020-08-04
Add blockchain.transaction.get_confirmed_blockhash

### LINKED ISSUE #287: Using electrs to emulate txindex
I'd like to write some kind of RPC proxy that uses electrs to emulate `txindex=1`. The idea is to forward all calls to `bitcoind` except for `getrawtransaction` without the `blockhash` parameter. That one would first find the height from `electrs` database, then get the blockhash from `bitcoind` and then forward the original request with the block hash parameter added.

This would be useful for people who want to run `electrs` and have `txindex` but don't want to store the index twice. It'd also help people who synchronized `bitcoind` and `electrs` without `txindex` and now they found out they need `txindex` and don't want to experience downtime from reindexing.

`electrs` database contains all information needed for this, however, it's not accessible over RPC.

I propose either of these solutions:

* Expose a new RPC method (e.g. `blockchain.transaction.get_height`) which translates txid to height. This is minimal change and all the logic has to be handled by external program.
* Expose a new (optional) listener that would implement all the logic described above. This would be more performant but with the increased maintenance burden.

What do you think? If you approve any of those solutions, I will definitely write a PR.

Kixunil:
This adds a new RPC method that enables users to fetch the block hash
for given txid. It can thus be used to emulate `txindex=1` as explained
in the related issue.

Closes #287

### DIFF
--- a/src/query.rs
+++ b/src/query.rs
@@ -431,6 +431,11 @@ impl Query {
             .gettransaction_raw(tx_hash, blockhash, verbose)
     }
 
+    pub fn get_confirmed_blockhash(&self, tx_hash: &Txid) -> Result<Value> {
+        let blockhash = self.lookup_confirmed_blockhash(tx_hash, None)?;
+        Ok(json!({ "block_hash": blockhash }))
+    }
+
     pub fn get_headers(&self, heights: &[usize]) -> Vec<HeaderEntry> {
         let _timer = self
             .duration
--- a/src/rpc.rs
+++ b/src/rpc.rs
@@ -267,6 +267,11 @@ impl Connection {
         Ok(self.query.get_transaction(&tx_hash, verbose)?)
     }
 
+    fn blockchain_transaction_get_confirmed_blockhash(&self, params: &[Value]) -> Result<Value> {
+        let tx_hash = hash_from_value(params.get(0)).chain_err(|| "bad tx_hash")?;
+        self.query.get_confirmed_blockhash(&tx_hash)
+    }
+
     fn blockchain_transaction_get_merkle(&self, params: &[Value]) -> Result<Value> {
         let tx_hash = hash_from_value(params.get(0)).chain_err(|| "bad tx_hash")?;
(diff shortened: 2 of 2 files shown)


romanz:
Thanks (and sorry for the delayed response)!

src/query.rs
```diff
@@ -431,6 +431,11 @@ impl Query {
             .gettransaction_raw(tx_hash, blockhash, verbose)
     }
 
+    pub fn get_confirmed_blockhash(&self, tx_hash: &Txid) -> Result<Value> {
+        let blockhash = self.lookup_confirmed_blockhash(tx_hash, None)?;
+        Ok(json!({ "block_hash": blockhash }))
+    }
```
Bit24888, line 437: ****

src/rpc.rs
```diff
@@ -267,6 +267,11 @@ impl Connection {
         Ok(self.query.get_transaction(&tx_hash, verbose)?)
     }
 
+    fn blockchain_transaction_get_confirmed_blockhash(&self, params: &[Value]) -> Result<Value> {
```
Bit24888, line 270: Screenshot_20200918-234855
Screenshot_20200918-234855

src/query.rs
```diff
@@ -431,6 +431,11 @@ impl Query {
...
+
```
Bit24888, line 438: Screenshot_20200918-234855

src/rpc.rs
```diff
@@ -318,6 +323,7 @@ impl Connection {
             "blockchain.transaction.broadcast" => self.blockchain_transaction_broadcast(&params),
             "blockchain.transaction.get" => self.blockchain_transaction_get(&params),
             "blockchain.transaction.get_merkle" => self.blockchain_transaction_get_merkle(&params),
+            "blockchain.transaction.get_confirmed_blockhash" => self.blockchain_transaction_get_confirmed_blockhash(&params),
```
Bit24888, line 326: C

PR #299, opened 2020-09-18
docker: account for host/container permission differences

jamesob:
Recently when running `docker build`, I encountered a cryptic error during `electrs` compilation (something having to do with `src/bulk.rs`) that I eventually figured out had to do with file permissions. This might be because I run with `umask 077` on my host, or it might be due to some kind of host/container uid mismatch (though my host's uid is 1000...). 

Anyway, this cautionary `chown` has fixed the build for me.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -8,9 +8,11 @@ RUN apt-get update \
 
 RUN adduser --disabled-login --system --shell /bin/false --uid 1000 user
 
-USER user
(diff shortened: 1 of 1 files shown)


romanz:
Thanks!

Kixunil:
I think adding `--home /home/user` to `adduser` would be cleaner, don't you think?

PR #301, opened 2020-09-28
Set Rust toolchain to 1.44.0

D4nte:
The project does not compile on 1.34.0 because it uses the lockfile
format introduced in 1.38.0.
There are also issues encountered with versions 1.39.0 to 1.43.0 as
seen here: https://github.com/D4nte/electrs/runs/1174430513

Do note that ubuntu-latest at this time uses Rust 1.46.0.

The oldest compatible version is 1.44.0. Setting a specific toolchain
version is good practice for binary project and remove potential issues
of incompatible toolchain for users trying to build the project.


Kixunil:
`electrs` is supposed to support 1.34 which is in Debian stable. The correct fix is to fix the lock file not to update the version.

romanz:
Many thanks for the PR!
IIUC, it seems that Debian stable now supports Rust 1.41.1: https://packages.debian.org/stable/rustc
I'll update the lock file :)

Kixunil:
Oh, I noticed some discussions about upgrading but didn't notice it's out already, such a great news! I agree with the update of the lock file.

D4nte:
@romanz as you can see here: https://github.com/D4nte/electrs/runs/1174430573

The issue with 1.41.0 compatibility is not the lock file but the use of some inner attributes:

```
error[E0658]: non-builtin inner attributes are unstable
 --> /home/runner/.cargo/registry/src/github.com-1ecc6299db9ec823/protobuf-2.17.0/src/well_known_types/wrappers.rs:9:1
  |
9 | #![rustfmt::skip]
  | ^^^^^^^^^^^^^^^^^
  |
  = note: for more information, see rust-lang/rust/issues/54726
```

I can update this PR assuming that 1.41.1 is to be used (and that it is working) and then later rebase once the issue above is fixed.

D4nte:
The issue is coming from `protobuf` which is a transitive dependency of `prometheus`.

Starting to get out of my depth here. Do you even use the protobuf part? It does not look like it can be feature gated unfortunately on prometheus side.

What is prometheus used for?

Only viable strategy would seem to downgrade prometheus to bring a compatible protobuf version in.

romanz:
`protobuf` was added due to `prometheus` - fixed at https://github.com/romanz/electrs/commit/aef1347859c2e0c98e3cc50bb8d07ba7ebfd742b.

romanz:
Merged manually at https://github.com/romanz/electrs/commit/3c4cf729600e94f83ee3104adf797306e10f77ba.

romanz:
Thanks!

romanz:
> What is prometheus used for?

It is used mainly for performance monitoring :)

Issue #308, opened 2020-10-24
electrs improvement proposal


romanz:
# Current issues

The current version of electrs has several open issues:

## RPC overhead

electrs issues RPCs to bitcoind to index the blocks' data (after the initial indexing is over).

In addition, it uses bitcoind to verify that the returned transactions are in the valid chain.

## Index structure

The current version uses multiple [indices](https://github.com/romanz/electrs/blob/master/doc/schema.md):

- scrtipthash -> funding txid
- funding txid -> spending txid
- txid -> confirmed height

So every scripthash query requires multiple RocksDB reads.

Also, the resulting DB takes 65 GB (as of Oct. 2020).

## Inconsistent history

When returning address status, the transactions [need to be ordered according to their appearence in a block](https://github.com/romanz/electrs/issues/297#issuecomment-695040748).

## Other

The current version doesn't use the latest Rust features (e.g. `async`) and probably can benefit from updating its dependencies (e.g. [RocksDB](https://github.com/rust-rocksdb/rust-rocksdb) and [rust-bitcoincore-rpc](https://github.com/rust-bitcoin/rust-bitcoincore-rpc/)).

# Proposal

By utilizing the "block undo" data (as done in [#14053](https://github.com/bitcoin/bitcoin/pull/14053) and [#14121](https://github.com/bitcoin/bitcoin/pull/14121), we can add [a new `getblocklocations` RPC to bitcoind](https://github.com/bitcoin/bitcoin/compare/master...romanz:locations).

It would allow electrs to get the actual file system locations of bitcoind [Block](https://github.com/bitcoin/bitcoin/blob/d67883d01e507dd22d1281f4a4860e79d6a46a47/src/primitives/block.h) and [BlockUndo](https://github.com/bitcoin/bitcoin/blob/d67883d01e507dd22d1281f4a4860e79d6a46a47/src/undo.h) data structures.

By reading this data directly from the file system, it is possible to directly build the following index:

| Script Hash Prefix   | `blk*.dat` file index | `blk*.dat` file offset |
| -------------------- | --------------------- | ---------------------- |
| `SHA256(script)[:8]` | 16-bit integer        | 32-bit integer         |

Today, Bitcoin mainnet has ~2300 `blk*.dat` files, each taking ~128MB.

This index allows retrieving an address' history using a single RocksDB read, followed by file system reads (which can be done concurrently if the blocks are stored on an SSD).

[...]

PR #315, opened 2020-11-04
Make rocksdb compression algos optional

Kixunil:
This makes the compression algorithms optional, but on by default to
keep the behavior. While one might think this just decreases the size a
bit, it has another advantage: when linking with system library it's not
needed to specify the paths of all these libraries because they are already
compiled in the system library.

Note: I have WIP update of the docs that will explain how to link to system
libraries, including with cross-compiling, I'm just waiting for some backport in
`bindgen`.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -18,6 +18,9 @@ spec = "config_spec.toml"
 [profile.release]
 lto = true
 
+[features]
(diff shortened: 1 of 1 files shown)


romanz:
Merged, thanks!

romanz:
Cherry-picked into `next` via 0f9ee7f30196eeb4d15439ac01976c3dfb5523da

romanz:
Could you please test the new electrs WIP version?
It can be built using the instructions here:
https://github.com/romanz/electrs/issues/308
I hope it should resolve this issue :)

PR #325, opened 2020-11-14
Updated librocksdb-sys and bindgen

Kixunil:
These new versions contain backports of fixes that make dynamic linking
less painful and cross compilation possible.

(I have a WIP doc update which I'm going to test now to see if the instructions work and then I'll file a PR.)


romanz (approved):
LGTM, thanks!

PR #327, opened 2020-11-15
Add simple health check to Dockerfile.

schildbach:
The check is based on the presence and responsiveness of the monitoring port 4224.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,7 @@ FROM rust:1.44.1-slim-buster
 
 RUN apt-get update \
   && apt-get install -y --no-install-recommends clang=1:7.* cmake=3.* \
-     libsnappy-dev=1.* \
+     libsnappy-dev=1.* curl \
   && apt-get clean \
   && rm -rf /var/lib/apt/lists/*
 
@@ -23,3 +23,5 @@ EXPOSE 50001
 EXPOSE 4224
 
 STOPSIGNAL SIGINT
+
+HEALTHCHECK CMD curl -fSs http://localhost:4224/ || exit 1


Dockerfile
```diff
@@ -23,3 +23,5 @@ EXPOSE 50001
 EXPOSE 4224
 
 STOPSIGNAL SIGINT
+
+HEALTHCHECK CMD curl -fSs http://localhost:4224/ || exit 1
```
romanz, line 27: Consider issuing an RPC to 50001 e.g. using:
```
contrib/health_check.py localhost 50001
```

romanz (changes requested):
LGTM - small suggestion below.

Dockerfile
schildbach: I considered it, but the problem is port 50001 isn't opened until all sync is finished. Which took around 16 hours in my case. 16 hours of unhealthy state is confusing I'd say.

Also, port 50001 is actually port 51001 on testnet, while port 4224 will stay the same afaik.

romanz: > 16 hours of unhealthy state is confusing I'd say.

Agreed.

romanz: The monitoring port (4224) is changed in Testnet to 14224:
https://github.com/romanz/electrs/blob/5510740a02e4f8140afd2de9a0a5d14c4f1b4eff/src/config.rs#L213

However, we can override it using:
 https://github.com/romanz/electrs/blob/5510740a02e4f8140afd2de9a0a5d14c4f1b4eff/config_spec.toml#L71

PR #333, opened 2020-11-24
Update lru to 0.6.1

Kixunil:
The old version of `lru` uses `std::mem::uninitialized()` which is UB
and actually panicks in the tests when compiled with Rust 1.48. (1.48
intentionally turned this UB to panicking.)

While 0.6.1 is breaking change according to semver rules it turns out no
code change was needed, just version bump.


romanz:
Many thanks!

PR #330, opened 2020-11-23
Run `cargo install` with  `--locked` in `Dockerfile`

Kixunil:
This prevents `cargo` from changing the dependencies during build.

Closes #329

PR #334, opened 2020-11-24
Update prometheus to 0.10

Kixunil:
The older version of `prometheus` depended on an old version of `spin`
which had a vulnerability related to locking. The new version of
prometheus doesn't use `spin` anymore so that the vulnerability is
fixed. Note that the vulnerability was unlikely on `x86_64` machines but
likely on ARM. `electrs` seems to be quite popular on various SBCs such
as Raspberry Pi, so fixing this is important.

More information: https://rustsec.org/advisories/RUSTSEC-2019-0013

Thankfully, `electrs` compiles with new `prometheus` without changes
despite it being semver breaking.

As you may have figured out, I ran `cargo audit` and it found one more vulnerability that to my knowledge is unrelated to `electrs` and a few smaller issues which I will sort out in a different PR (#335).

@romanz I'd strongly recommend releasing a new version with this fix, ideally with #333 and #330 as well to avoid people spamming with "electrs does not compile" issues.

Pinging maintainers of various SBC projects: @rootzoll @stadicus @openoms @ketominer @lukechilds (please CC if I forgot someone).

## Is my project vulnerable?

* If you didn't expose prometheus port and you assume all users on your machine are trusted or you isolated prometheus interface somehow and didn't expose it, then the vulnerability *might* not be present (this is UB, so it's impossible to say for sure)
* If you filter RPC methods available to `electrs` using Bitcoind 0.20 whitelist feature or using [`btc-rpc-proxy`](https://github.com/Kixunil/btc-rpc-proxy/) then at worst the incoming money could be faked after successful exploitation
* I guess exploiting this vulnerability is not easy as it likely relies on scheduling of threads which is not deterministic

Regardless, upgrading once a new version is released is definitely a good idea.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -81,7 +81,7 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "46254cf2fdcdf1badb5934448c1bcbe046a56537b3987d96c51a7afc5d03f293"
 dependencies = [
  "addr2line",
- "cfg-if",
+ "cfg-if 0.1.10",
  "libc",
  "miniz_oxide",
  "object",
@@ -127,7 +127,7 @@ checksum = "2a817f8356b784c37ee29b7a9a15c3fec321cd46b0ec67bcebe5530207f62e2d"
 dependencies = [
  "bitflags",
  "cexpr",
- "cfg-if",
+ "cfg-if 0.1.10",
  "clang-sys",
  "clap",
  "env_logger",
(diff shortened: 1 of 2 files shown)


rootzoll:
We are preparing a RaspiBlitz 1.6.2 release at the moment ... is there a time estimation on when this fix will be part of an electrs release so we might wait for it?

Kixunil:
I can't speak for @romanz but the fixes are trivial. Probably doesn't hurt to wait a bit if you're not in a hurry.

romanz:
Many thanks for the contribution and sorry for the delayed response.

romanz:
The release will be ready soon :)

romanz:
A new release is ready: https://crates.io/crates/electrs/0.8.6
Many thanks for reporting and investigating this issue!

PR #335, opened 2020-11-24
Replace unmaintained crates with alternatives

Kixunil:
According to `cargo audit` crates `rust-crypto`, `dirs` and `failure`
are unmaintained/deprecated. This change replaces former two.

Unfortunately `failure` can't be replaced because it's a transitive
dependency of `rocksdb`, which is pinned. Fortunately it's in build
script only, which should have no noticable effect on the quality of
`electrs` itself.

As a nice side effect this should remove a bunch of useless code since
we pull only specifically `sha2` without bundling other cryptography.


romanz:
Thanks!

PR #345, opened 2020-12-01
Added upgrade instructions

Kixunil:
This adds upgrade instructions as well as a note about RAM usage during
build which was missing.


romanz (approved):
Many thanks, much appreciated!

PR #358, opened 2021-02-01
Updated several dependencies

### LINKED ISSUE #354: Feature: Update configure_me(_codegen) to 0.4
`configure_me` 0.4 updated `toml` to 0.5. There's no significant breakage but the new version no longer accepts some configuration files that were invalid Toml - when they contain headers twice. This is not a problem for `electrs` because it never supported such configuration. The biggest advantage is de-duplicating `toml` dependency - there are currently two versions, 0.4 and 0.5 (coming from `cargo_toml`).

Update dependencies.

Use ancient `toml` crate and keep duplicated dependencies.

**Additional context**
I can make a PR later, but don't have time to test it properly right now.

Kixunil:
* `arc_swap` and `smallvec` fix security vulnerabilities
* `configure_me` and `configure_me_codegen` update toml dependency
  leading to deduplication and shorter compile times
* `log` upgrades dependency on `cfg-if` bringing us closer to
  deduplicating it.

Closes #354

I intend to update more things in separate PR(s). This one should be low-hanging, safe and important stuff.
I will try to hurry up so that we can release a new version if those vulnerabilities actually affect `electrs` (didn't check yet).


Kixunil:
Changed my mind and added `stderrlog` into this PR to reduce noise. There's not much more stuff. Just `rocksdb` is blocking us from other huge deduplications and `prometheus` could update its own dependencies but it's their choice.

romanz:
Thanks!

Issue #344, opened 2020-11-30
Help: How to update electrs?


jonjonez:
How to update electrs?

I'm having problems with Electrum 4.0.5. how can I update the electrs server to version 0.8.6? I saw in the recent issues that this new update solves the problem of connectivity with Electrum.

However I didn't find any upgrade manual in the repository guide, and the ones I found on google didn't work for me.

Can someone help me?

jonjonez:
Whenever I try to update, I get this feedback:
HEAD is now in version 5bb6f09 Bump

Link image: https://imgur.com/qYNyRqQ

Kixunil:
Nothing fancy, just replace the binary and restart.

You're reading a guide for Raspberry Pi but not running on Raspberry Pi, so obvioudly the instructions don't work. The [official `electrs` documentation](https://github.com/romanz/electrs/blob/master/doc/usage.md) has very detailed instructions on how to build `electrs`.

If you use Debian (or derived distro), then just `apt install cargo` works. However I wonder how did you even compile `electrs` without Rust in the first place?

jonjonez:
> Nothing fancy, just replace the binary and restart.
> 
> You're reading a guide for Raspberry Pi but not running on Raspberry Pi, so obvioudly the instructions don't work. The [official `electrs` documentation](https://github.com/romanz/electrs/blob/master/doc/usage.md) has very detailed instructions on how to build `electrs`.
> 
> If you use Debian (or derived distro), then just `apt install cargo` works. However I wonder how did you even compile `electrs` without Rust in the first place?

When trying to install a new version of rust, the terminal returns to me that a new version is already installed, when giving the command: 'apt install cargo' it does not let me install.

I'm sorry if that sounds too basic, but I'm not really able to understand the best way to update.

My linux is Ubuntu 18.04

Kixunil:
That should be fine, Ubuntu 18.04 has sufficient Rust version for compiling `electrs`. Try to build it with `cargo build --release` and if no error pops out, you'll find the binary in `./target/release/electrs`

You might be interested in [my project](https://deb.ln-ask.me) which intends to solve various things like how to configure these apps, how to upgrade... Beware though that you would have to somehow migrate your existing setup to take full advantage of it.

jonjonez:
Your project is very interesting, I will keep an eye on it, I found it very promising, hopefully it will revenge and mature...

Anyway, I tried to do it now and I get the same mistakes, honestly I don't know what to do, I've tried everything.

[...]

PR #359, opened 2021-02-01
Improve doc

Kixunil:
This improves documentation of configuration based on @BTCBellyButton
suggestions in #344. It contains a few more tweaks.

I have also added a few more line breaks to make editing easier (separate commit to simplify review).


romanz:
Thanks a lot!

PR #360, opened 2021-02-01
Implemented auth option and deprecated cookie

### LINKED ISSUE #349: Feature: rename cookie to auth (or something similar)
The `cookie` name is confusing as currently it should be used for `username:password`. Putting actual cookie there is a bad idea since bitcoind changes it on every restart. So every restart would make it invalid and force restart of electrs. Thus cookie is only useful for password authentication.

For the next at least one version start accepting both `cookie` and `auth` parameters. If `cookie` is given, print out a deprecation warning regardless of verbosity. Then some time in the future, make it a hard error, but keep detecting it for a while. Then remove it completely.

When both are given, it's a hard error.

If an emergency release of `electrs` is released very soon after previous version then we extend the period to give the users enough time. This implies that changing from warning to hard error and removing needs to be timed right before a release.

* Keep calling it cookie but improve the docs - I'm not convinced this is sufficient and it makes the docs longer (it's quite long already)
* Various alternatives of keeping the option indefinitely etc. - I believe that the approach I described is a good compromise. We (developers) don't want extra code for handling historical stuff but the users need to be notified well-enough. At first it will be about turning off a warning, the hard error period is designed for people who run `electrs --conf ... &>/dev/null` as those would never see the warning otherwise. (We could write to the console, but that's gross.) Last stage removes all the code related to `cookie`.

**Additional context**

Kixunil:
This adds `auth` option and deprecates `cookie` which has a confusing
name. It's effectively renaming but we pretend it is not to avoid people
blindly changing the name without second thought.

Closes #349 (a new tracking issue will be created)

This is WIP because I intend to update the doc too but not before #359 is merged.

### DIFF
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -219,13 +219,18 @@ You need to use a number in config file if you want to increase verbosity (e.g. 
 
 **Authentication**
 
-In addition, config files support `cookie` option (warning: it will be renamed in near future!) to specify username and password.
-This is not available using command line or environment variables for security reasons (other applications could read it otherwise).
-**Important note**: `cookie` is different from `cookie_file`, which points to a file containing the cookie instead of being the cookie itself!
+**Warning:** If you're reading this before version 0.8.8 exists either use `cookie_file` or check the version of documentation before this line was committed.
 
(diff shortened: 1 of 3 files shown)


Kixunil:
I have documented deprecation. Will be good to release a new version soonish to not confuse people too much. I'd still like to wait for `prometheus` update though.

romanz (approved):
LGTM

romanz:
Thanks!

PR #361, opened 2021-02-04
Use sync primitives from parking_lot

Kixunil:
This change adds a custom module mimicking `std::sync` which may contain
custom implementations of synchronization primitives. It also redirects
all cases of `std::sync` to use this new module and replaces mutexes
with thosse from `parking_lot`.

The main motivation is to help with finding deadlocks using deadlock
detection feature from `parking_lot`. The locks from `parking_lot` don't
implement poisoning so we needed to add it on top to maintain perfect
compatibility. This was solved by wrapping locks in custom types which
mirror the `std` API.

Deadlock detection is behind a feature flag to not slow donw production
use and also because it's considered experimental.

Because this change makes swapping primitives easy, we could try
replacing more of them in the future and measure performance impact.
However I don't expect significant changes and we should focus on
low-hanging fruit first.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -369,6 +369,7 @@ dependencies = [
  "lru",
  "num_cpus",
  "page_size",
+ "parking_lot",
  "prometheus",
  "protobuf",
  "rocksdb",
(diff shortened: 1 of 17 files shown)


tcharding (commented):
Nice idea. This whole `sync` module reads like boilerplate, could this not be in a crate of its own and we just depend upon it?

Kixunil:
Perhaps the implementation of poisoning mutexes could/should. The rest is redirection specific to `electrs` so it has to be in its own module.

tcharding:
> Perhaps the implementation of poisoning mutexes could/should. The rest is redirection specific to `electrs` so it has to be in its own module.

Oh, my bad, I missed that. Thanks

Kixunil:
No problem, I actually started to like the idea of moving it into separate library very much. It will both look cleaner and has the opportunity to help others.

romanz:
It will be implemented in [0.9](https://github.com/romanz/electrs/milestone/1).

PR #372, opened 2021-02-22
Added in-depth explanation of cookie deprecation

Kixunil:
This adds more informtion about cookie deprecation, especially
motivation, so the users won't be frustrated because of unexplained
change. It also removes a warning that's no longer applicable.


romanz (commented):
LGTM, thanks!

PR #378, opened 2021-02-28
[ZW-4847] Custom transaction speed - estimatefee to use Bitcoin-Core's estimatesmartfee

oleiba:


.github/workflows/rust.yml
```diff
@@ -9,6 +9,11 @@ jobs:
 
     steps:
     - uses: actions/checkout@v1
+    - name: Install 1.41.0-nightly (3e525e3f6 2019-11-18)
```
Kixunil, line 12: Why use nightly?

.gitignore
```diff
@@ -5,3 +5,5 @@ _*/
 *.sublime*
 *~
 *.pyc
+.idea
+.idea/*
```
Kixunil, line 9: I believe it's better to put editor-specific gitignores to `~/.gitignore`, however I see sublime there too...

Cargo.toml
```diff
@@ -48,6 +48,8 @@ stderrlog = "0.4.1"
 sysconf = ">=0.3.4"
 time = "0.1"
 tiny_http = "0.6"
+rusoto_core = "0.36.0"
+rusoto_dynamodb = "0.36.0"
```
Kixunil, line 52: Why should `electrs` depend on AWS? I don't use AWS, I don't want this dependency.
But it's even worse, Rusoto says: "Rusoto is in maintenance mode."

config_spec.toml
```diff
@@ -93,11 +93,17 @@ type = "f32"
...
+type = "usize"
+doc = "Number of transactions to lookup before returning a warning that the amount is getting close to the true limit; See txid_limit below"
+default = "250"
+
 [[param]]
 name = "txid_limit"
 type = "usize"
 doc = "Number of transactions to lookup before returning an error, to prevent 'too popular' addresses from causing the RPC server to get stuck (0 - disable the limit)"
-default = "100"
+default = "0"
```
Kixunil, line 106: Can you explain this change?

config_spec.toml
```diff
@@ -93,11 +93,17 @@ type = "f32"
 doc = "Total size of block transactions IDs to cache (in MB)"
 default = "10.0"
 
+[[param]]
+name = "txid_warning_limit"
+type = "usize"
+doc = "Number of transactions to lookup before returning a warning that the amount is getting close to the true limit; See txid_limit below"
+default = "250"
+
```
Kixunil, line 101: What is the use of this warning? If an address has too many transactions, the query will fail, if it hasn't it will succeed.

Kixunil (commented):
Could you please describe your changes in more detail? What are you trying to achieve? Why did you write it this way? It's very hard to review the code without knowing these.

BTW I'm quite surprised by this change out-of-blue without any explanation or previous issue. I don't think this is a very efficient approach.

Anyway, the final decision should be done by @romanz, I'm mainly stating my opinions.

oleiba:
I am sorry, this was opened by mistake and was supposed to be merged to the master of a different account.

Kixunil:
Ah, OK, feel free to contribute if you have something useful! :)

PR #379, opened 2021-03-03
Fixed inverted logic of deprecation

Kixunil:
The warning was supposed to be shown when `cookie` is specified not the
other way around.

I have no clue how this slipped through, sorry. :(


romanz:
No worries, thanks for the fix!

Issue #382, opened 2021-03-13
Heavy Electrum wallets wont synchronize.


ascisco:
Is there a way to configure electrs to handle heavy wallets (wallets with a lot of transactions in them)? After my wallet has a 100 or more transactions in it, it starts get stuck and hangs on Synchronizing... 0/1 in my electrum wallet. 

Im running the latest raspiblitz and electrum mobile wallet.

Kixunil:
Sadly this is a known issue that was repeated here multiple times, I'm trying to fix it but it goes slowly since I don't have that lot of time. But based on lot of feedback like this, I will try to prioritize it.

craigraw:
May I suggest that in the meantime there is a new minor release which sets the `wait_duration_secs` configuration setting from 5 to 10 seconds? This quick fix would relieve this issue for many people while the actual fix is coming.

Kixunil:
@craigraw didn't notice bumping it helps, thanks for info! I opened #384.

There are other valuable fixes in the master, so I agree minor release would be useful.

romanz:
0.8.9 is [released](https://crates.io/crates/electrs).

romanz:
Please see https://github.com/romanz/electrs/issues/308#issuecomment-806061229

PR #384, opened 2021-03-18
Bumped default wait_duration_secs from 5 to 10

Kixunil:
According to @craigraw this should help with performance issues while a
more complicated improvement is being developed.

See #382


romanz:
Thanks!

PR #387, opened 2021-03-21
Optimize Dockerfile

### LINKED ISSUE #322: Dockerfile seems to be missing a CMD
Usually Dockerfiles are self contained, so you can run them and the included daemon process starts on itself. This last part seems to be missing, which is confusing.

Code1110:
This adds an Entrypoint to the Dockerfile and fixes #322.

It also introduces multistage Docker build, that reduces the image size from ~1.85GB to ~463MB, because the rust toolchain is no longer in the resulting image.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,20 +1,29 @@
-FROM rust:1.44.1-slim-buster
+FROM rust:1.44.1-slim-buster as builder
 
 RUN apt-get update \
   && apt-get install -y --no-install-recommends clang=1:7.* cmake=3.* \
-     libsnappy-dev=1.* curl \
+  libsnappy-dev=1.* curl \
   && apt-get clean \
   && rm -rf /var/lib/apt/lists/*
 
 RUN adduser --disabled-login --system --shell /bin/false --uid 1000 user
 
 WORKDIR /home/user
-COPY ./ /home/user
+COPY . .
 RUN chown -R user .
 
-USER user
-
 RUN cargo install --locked --path .
+
+# Create runtime image
+FROM debian:buster-slim
+
+RUN adduser --disabled-login --system --shell /bin/false --uid 1000 user
+
+WORKDIR /home/user/app
+RUN chown user .
+COPY --from=builder /home/user/target/release .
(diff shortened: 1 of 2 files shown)


romanz (approved):
LGTM

romanz:
Thanks!

Kixunil:
I wonder what costs/benefits would be using dynamic linking. It should at least provide faster compile time.

Is dynamic linking usual when Docker is used?

Code1110:
Hmm, I can't answer this question, because I'm not familiar with Rust. I only know that it takes around 7 mins. to build the image, which is pretty long imho.

Kixunil:
Dynamic linking against C system libraries is same in Rust as in C/C++. 7 mins is pretty good time for `electrs` build, it usually takes me far more than that with static linking. If I remember correctly dynamic linking cuts the time roughly in half.

The question boils down to "Is it usual for applications in Docker container to depend on things installed via apt inside that container?" Is updating via apt even a thing in Docker?

If this is a common or at least OK practice and those 7 min is considered large, then I recommend switching to dynamic linking.

Code1110:
It's OK to install dependencies for an application using apt. In general, a Docker container is considered immutable, so when dependencies change/are updated the Dockerfile should be changed, and a new image should be created. The images can be versioned using tags, e.g. `electrs:v0.8.9`.

Kixunil:
Great! Then I don't think there's anything blocking dynamic linking. Should be easy to use.

>  a new image should be created.

WTF, does it update the whole thing just because a library updated?

Code1110:
Not necessarily. Each command in the Dockerfile creates a new layer, that is cached and reused across different images.
It really depends on the use case. If you want to use a container just for yourself, you can go inside that container and do an apt update. If you want to provide prebuild images for others to use, you probably want to create a new image for every release.

Kixunil:
Suppose in theory that a vulnerability is found in rocksdb. A user uses public Docker image of `electrs` which gets updated. Will this cause this user to re-download the whole image or just the changed part?

PR #390, opened 2021-04-02
Fix #313

### LINKED ISSUE #313: JSON-RPC errors are sent as string instead of object, not following JSON-RPC spec
Protocol errors are not sent correctly to the client. See https://www.jsonrpc.org/specification#error_object

-----

Example:
when I connect to Electrs and subscribe to a large history address, I get this message:
`{'error': 'Too many history entries', 'id': 11, 'jsonrpc': '2.0'}`
Electrum (the client) will then log:
```
I/i | interface.[cryptobadger.io:50002] | connection established. version: ['electrs 0.4.1', '1.4']
[...]
E/i | interface.[cryptobadger.io:50002] | Exception in wrapper_func: ProtocolError(-32600, 'ill-formed response error object: Too many history entries')
Traceback (most recent call last):
  File "...\electrum\electrum\util.py", line 1039, in wrapper
    return await func(*args, **kwargs)
  File "...\electrum\electrum\interface.py", line 480, in wrapper_func
    return await func(self, *args, **kwargs)
  File "...\electrum\electrum\interface.py", line 504, in run
    await self.open_session(ssl_context)
  File "...\electrum\electrum\interface.py", line 644, in open_session
    await group.spawn(self.monitor_connection)
  File "...\aiorpcX\aiorpcx\curio.py", line 242, in __aexit__
    await self.join()
  File "...\aiorpcX\aiorpcx\curio.py", line 211, in join
    raise task.exception()
  File "...\electrum\electrum\synchronizer.py", line 81, in _start_tasks
    await group.spawn(self.main())
  File "...\aiorpcX\aiorpcx\curio.py", line 242, in __aexit__
    await self.join()
  File "...\aiorpcX\aiorpcx\curio.py", line 211, in join
    raise task.exception()
  File "...\electrum\electrum\synchronizer.py", line 109, in subscribe_to_address
    await self.session.subscribe('blockchain.scripthash.subscribe', [h], self.status_queue)
  File "...\electrum\electrum\interface.py", line 194, in subscribe
    result = await self.send_request(method, params)
  File "...\electrum\electrum\interface.py", line 170, in send_request
    response = await asyncio.wait_for(
  File "...\Python38\lib\asyncio\tasks.py", line 455, in wait_for
    return await fut
  File "...\aiorpcX\aiorpcx\session.py", line 537, in send_request
    return await self._send_concurrent(message, future, 1)
  File "...\aiorpcX\aiorpcx\session.py", line 507, in _send_concurrent
    return await future
  File "...\aiorpcX\aiorpcx\jsonrpc.py", line 721, in receive_message
    item, request_id = self._protocol.message_to_item(message)
  File "...\aiorpcX\aiorpcx\jsonrpc.py", line 273, in message_to_item
    return cls._process_response(payload)
  File "...\aiorpcX\aiorpcx\jsonrpc.py", line 220, in _process_response
    raise cls._error(code, message, False, request_id)
aiorpcx.jsonrpc.ProtocolError: (-32600, 'ill-formed response error object: Too many history entries')
```

If I do the same but connect to ElectrumX, I get:
`{'jsonrpc': '2.0', 'error': {'code': 1, 'message': 'history too large'}, 'id': 11}`
Electrum (the client) will then log:
```
I/i | interface.[fortress.qtornado.com:443] | connection established. version: ['ElectrumX 1.15.0', '1.4']
[...]
E/i | interface.[fortress.qtornado.com:443] | disconnecting due to GracefulDisconnect(RPCError(1, 'history too large'))
```

romanz:


Kixunil (approved):
Looks good.

Something to think about in the future: return generic error and log the real error to avoid potential privacy leaks.

knocte:
Hardcode code=1? why? sorry, this doesn't look good to me. I think code=1 is already taken, check electrumX implementation.

Kixunil:
@knocte ah, good point! The error codes are [not documented](https://electrum.readthedocs.io/en/latest/protocol.html). I'm not sure if ElectrumX source code is authoritative specification but if there's nothing better...

knocte:
Look, am not claiming or suggesting that ElectrumX is authoritative in any way, but given that electrs would introduce for the first time error codes, it only makes sense to try to adapt to what other implementations (which have had error codes for a long time) do. So I'm just talking for the sake of simplicity, and compatibility, from the point of view of a client consuming these errors, so that we don't need to sniff and find out what server we're dealing with. Error codes allow us to avoid parsing error messages, but if error codes are going to be different or conflicting depending on the server implementation, then they don't help us at all.

Kixunil:
Alright, do the clients actually depend on those error codes today? Perhaps having "other error" for starters and improving it in the future is better than trying to fix everything in one go?

romanz:
IIUC, ElectrumX is using the following error codes:
```python
BAD_REQUEST = 1
DAEMON_ERROR = 2
```
https://github.com/spesmilo/electrumx/blob/557cb5821f09fc8ccecce09d215a0f94d216100e/electrumx/server/session.py#L47-L48

knocte:
Guys, it would be nice if you let me some time to reply :)

> ElectrumX is using the following error codes:

We have found many more than this, will follow-up soon. Can you hold off making a release for now at least? Thanks

romanz:
Sorry for the misunderstanding...
Will hold off making the release until the issue is resolved.

canndrew:
Based on my investigations, electrumx used to have a bunch more error codes but at some point they replaced their json-rpc implementation and changed all the error codes in the process. In the current version they have two application-defined error codes (`BAD_REQUEST` and `DAEMON_ERROR`, as linked above) but there's still the error codes that are part of the json-rpc spec. These are listed here: https://www.jsonrpc.org/specification#error_object

I think it would make sense to replace the json-rpc impl in electrs with either an existing json-rpc library or a minimal reimplementation living in its own module or crate, and, either way, to make sure the implementation is spec-compliant. In particular the error codes that are part of the spec should be generated in that module/library and the next layer up would be able to pass down its own error codes. Having the json-rpc implementation be separated from the rest of the code would also make sense since json-rpc is its own, well-defined thing.

We could then look at electrumx again and see if there's any consistency or logic behind where it uses `BAD_REQUEST` vs `DAEMON_ERROR` and try to follow suit.

Kixunil:
I took a look at various crates and so far [`json_rpc2`](https://docs.rs/json-rpc2/0.10.1/json_rpc2/) looks the best. It seems to be atively developed, supports server and has a proper error type.

We would need to bubble-down the error type to the places where we can distinguish the kinds of errors. I believe this is the right thing to do.

knocte:
Hey @romanz and @Kixunil, Andrew Cann has a WIP commit here: https://github.com/canndrew/electrs/commit/5f298bcf6d5e21de05436817b26cc3c483ce3885 do you guys like how he's approaching it or can we get an early review to maybe adjust his first steps?

Kixunil:
@knocte sadly I don't have the time in upcoming days to review that big diff. Just one thing to note: maybe it's more suitable for `next` branch, not sure. What do you think @romanz?

romanz:
Thanks @knocte and @canndrew!
If possible, I would appreciate if you could please rebase this work over the `p2p` branch: https://github.com/romanz/electrs/issues/308#issuecomment-806061229, which (hopefully) will become the 0.9.0 release.

canndrew:
So I'll explain what I'm thinking with this..

I had a look at the `json-rpc2` library but I have a couple of issues with it. Firstly it lacks type-safety in some places. Some examples: it's possible to return method responses that contain neither a success result nor an error, request ids aren't constrained to being integers or strings and method parameters aren't constrained to being arrays or objects. Since this issue started with electrs returning invalid JSON-RPC responses I think it's a good idea to try and strictly enforce the spec at the level of the JSON-RPC implementation. It also doesn't allow you to return custom error codes, which is necessary for mirroring electrumx's error codes. It also doesn't seem to allow you to handle notifications correctly. It expects your `Service` implementation to return a response for every request, or `None` if the request can be handled due to an invalid method, however not all JSON-RPC calls expect a response at all. 

Since the JSON-RPC spec is small enough to implement in a day I thought I'd have a crack at implementing it, and making sure the implementation enforces every word of the spec. That's pretty much done, but it needs some testing, commenting and general clean up before it can be considered for being integrated.

One thing to note though is that I implemented it in async rust on top of tokio, mainly because it was easier that way. The current electrs branch is using system threads for parallelism, however it would be possible to have an async runtime while still dispatching CPU-intensive work to worker threads. Alternatively I could re-work the implementation to use systems threads and channels. However I notice the `next` branch is using async-std, and porting from tokio to async-std would be trivial. What's the status of the `next` branch though? Is it currently usable and slated for release in the near-term?

romanz:
`next` branch is assuming an RPC for fetching block locations from Bitcoin Core for efficient indexing and queries:
https://github.com/bitcoin/bitcoin/pull/20702

Since this PR didn't get merged, I am now working on `p2p` branch which uses p2p-based sync from Bitcoin Core - which is much faster than of JSONRPC-based sync.

romanz:
Currently, https://github.com/romanz/electrs/blob/d442501e2677d510f71015cbfc133814e8005aa2/src/server.rs is using threads, but if it's easier/clearer to rewrite using Tokio - please let me know.

canndrew:
I think tokio would be easier/clearer. In my experience, having threads passing messages over channels tends to make code very goto-ish, in the sense that it's difficult to reason about control flow, harder to debug, and harder to make sure resources/tasks get cleaned up correctly (since they're not scoped under the code that created them). Tokio/async code tends to be more concise too. Do you have a preference between tokio and async-std?

romanz:
> Do you have a preference between tokio and async-std?

No - but please make sure that the relevant code/libraries compile with Rust 1.41.1 (current Rust version under stable Debian).

canndrew:
> No - but please make sure that the relevant code/libraries compile with Rust 1.41.1 (current Rust version under stable Debian).

Okay. It seems that the latest tokio version won't compile with 1.41, but there are earlier ones that will, and I'm not using any new features so downgrading shouldn't be much more work that just changing the version number.

I've done some more work on the RPC impl. I haven't rebased onto the p2p branch yet, instead I've expanded the test case that was there and started writing a derive-like macro for `JsonRpcService` implementations which handles the boilerplate of dispatching on methods, checking and parsing arguments and serializing results, and makes sure to return the correct spec-compliant error codes.

Kixunil:
Sorry guys for confusion I meant to write `p2p` not `next`.

Thanks as lot @canndrew for working on this!!! :heart:

romanz:
Sounds great, and many thanks for the contribution!

canndrew:
No probs :). I'm still working on it so I've pushed my WIP changes. I've implemented that macro, but it's buggy and generates code which fails type-checking. I'll keep working on it tomorrow and I'll expect I'll be able to get it working. The steps after that are: testing; clean up and document the code; rebase on top of p2p; integrate with the rest of electrs.

Issue #294, opened 2020-08-17
error compiling sysconf on FreeBSD 11.3


seth586:
```
Compiling sysconf v0.3.4
error[E0425]: cannot find value `_SC_XBS5_ILP32_OFF32` in crate `libc`
   --> /root/.cargo/registry/src/github.com-1ecc6299db9ec823/sysconf-0.3.4/src/raw.rs:85:28
    |
85  |     ScXbs5Ilp32Off32 = sc!(_SC_XBS5_ILP32_OFF32),
    |                            ^^^^^^^^^^^^^^^^^^^^ help: a constant with a similar name exists: `_SC_V6_ILP32_OFF32`
    |
   ::: /root/.cargo/registry/src/github.com-1ecc6299db9ec823/libc-0.2.65/src/unix/bsd/freebsdlike/mod.rs:913:1
    |
913 | pub const _SC_V6_ILP32_OFF32: ::c_int = 103;
    | -------------------------------------------- similarly named constant `_SC_V6_ILP32_OFF32` defined here

error[E0425]: cannot find value `_SC_XBS5_ILP32_OFFBIG` in crate `libc`
   --> /root/.cargo/registry/src/github.com-1ecc6299db9ec823/sysconf-0.3.4/src/raw.rs:86:29
    |
86  |     ScXbs5Ilp32Offbig = sc!(_SC_XBS5_ILP32_OFFBIG),
    |                             ^^^^^^^^^^^^^^^^^^^^^ help: a constant with a similar name exists: `_SC_V6_ILP32_OFFBIG`
    |
   ::: /root/.cargo/registry/src/github.com-1ecc6299db9ec823/libc-0.2.65/src/unix/bsd/freebsdlike/mod.rs:914:1
    |
914 | pub const _SC_V6_ILP32_OFFBIG: ::c_int = 104;
    | --------------------------------------------- similarly named constant `_SC_V6_ILP32_OFFBIG` defined here

error[E0425]: cannot find value `_SC_XBS5_LPBIG_OFFBIG` in crate `libc`
   --> /root/.cargo/registry/src/github.com-1ecc6299db9ec823/sysconf-0.3.4/src/raw.rs:87:29
    |
87  |     ScXbs5LpbigOffbig = sc!(_SC_XBS5_LPBIG_OFFBIG),
    |                             ^^^^^^^^^^^^^^^^^^^^^ help: a constant with a similar name exists: `_SC_V6_LPBIG_OFFBIG`
    |
[...]

Issue #132, opened 2019-04-10
Unable to compile on FreeBSD


seth586:
#### Environment:
FreeBSD 11.2 x64
clang --version:
```
root@bitcoin:~/electrs # clang --version
FreeBSD clang version 6.0.0 (tags/RELEASE_600/final 326565) (based on LLVM 6.0.0)
Target: x86_64-unknown-freebsd11.2
Thread model: posix
InstalledDir: /usr/bin
```
cargo --version:
```
root@bitcoin:~/electrs # cargo --version
cargo 1.33.0
```

#### Output `cargo build --release`
```
root@bitcoin:~/electrs # cargo build --release
   Compiling arrayvec v0.4.10
   Compiling libc v0.2.50
   Compiling nodrop v0.1.13
   Compiling cfg-if v0.1.7
   Compiling scopeguard v0.3.3
   Compiling memoffset v0.2.1
   Compiling lazy_static v1.3.0
   Compiling rayon-core v1.4.1
   Compiling rayon v1.0.3
   Compiling proc-macro2 v0.4.27
   Compiling either v1.5.1
   Compiling unicode-xid v0.1.0
   Compiling syn v0.15.29
   Compiling autocfg v0.1.2
   Compiling memchr v2.2.0
   Compiling failure_derive v0.1.5
   Compiling version_check v0.1.5
   Compiling rustc-demangle v0.1.13
   Compiling glob v0.2.11
   Compiling ucd-util v0.1.3
   Compiling regex v1.1.2
   Compiling byteorder v1.3.1
   Compiling unicode-width v0.1.5
   Compiling utf8-ranges v1.0.2
   Compiling quick-error v1.2.2
   Compiling bitflags v1.0.4
   Compiling strsim v0.7.0
   Compiling bindgen v0.47.3
   Compiling vec_map v0.8.1
   Compiling termcolor v1.0.4
   Compiling ansi_term v0.11.0
   Compiling num-traits v0.2.6
   Compiling peeking_take_while v0.1.2
   Compiling smallvec v0.6.9
   Compiling winapi-build v0.1.1
   Compiling matches v0.1.8
   Compiling num-integer v0.1.39
   Compiling protobuf v2.4.0
   Compiling serde v1.0.89
   Compiling ryu v0.2.7
   Compiling gcc v0.3.55
   Compiling winapi v0.2.8
   Compiling bit-vec v0.4.4
   Compiling prometheus v0.5.0
   Compiling bech32 v0.6.0
[...]

PR #395, opened 2021-04-19
Make sysconf dependency optional

wyager:
Allow building without sysconf dependency. Re. https://github.com/romanz/electrs/issues/294

### DIFF
--- a/src/metrics.rs
+++ b/src/metrics.rs
@@ -104,13 +104,24 @@ struct Stats {
     fds: usize,
 }
 
+#[cfg(feature="sysconf")]
+fn ticks_per_second() -> Result<f64> {
+    sysconf::raw::sysconf(sysconf::raw::SysconfVariable::ScClkTck)
+        .map_err(|_| "failed to get _SC_CLK_TCK".into())
+        .map(|x| x as f64)
+}
+
+#[cfg(not(feature="sysconf"))]
+fn ticks_per_second() -> Result<f64> {
+    Err("could not get _SC_CLK_TCK - not built with `--features sysconf`".into())
+}
+
 fn parse_stats() -> Result<Stats> {
     let value =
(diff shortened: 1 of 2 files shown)


wyager:
Also re. https://github.com/romanz/electrs/issues/132

src/metrics.rs
```diff
@@ -104,13 +104,24 @@ struct Stats {
     fds: usize,
 }
 
+#[cfg(feature="sysconf")]
+fn ticks_per_second() -> Result<f64> {
+    sysconf::raw::sysconf(sysconf::raw::SysconfVariable::ScClkTck)
+        .map_err(|_| "failed to get _SC_CLK_TCK".into())
```
Kixunil, line 110: This loses valuable error information that could be important for debugging. Please keep the error somehow (displaying it is OK).

src/metrics.rs
```diff
@@ -104,13 +104,24 @@ struct Stats {
...
+        .map(|x| x as f64)
+}
+
+#[cfg(not(feature="sysconf"))]
+fn ticks_per_second() -> Result<f64> {
+    Err("could not get _SC_CLK_TCK - not built with `--features sysconf`".into())
```
Kixunil, line 116: Since this causes `parse_stats` to unconditionally fail, maybe gate that one as well? Not sure how prometheus works, wouldn't this flood the log with error messages?

Kixunil (changes requested):
I have very bad experiences with throwing away errors, so I strongly recommend not to.

romanz:
It should be resolved by https://github.com/romanz/electrs/commit/8890c4bbb78d2de6a6f3e9e2b5ef8094106745a4 (in `p2p` branch).

romanz:
Soon to be released in `0.9.0-rc1` :)

PR #401, opened 2021-05-20
Updated sha2

Kixunil:
The new version of sha2 replaces its dependency `cpuid-bool` with
`cpufeatures` because it was renamed and the previous one is
unmaintained.


romanz:
Thanks!

PR #402, opened 2021-05-20
Updated tiny_http

Kixunil:
The previous version of `tiny_http` had a vulnerability that's probably
not relevant to `electrs` but better not risk it. It also avoids ugly
red error from `cargo audit`. :)


romanz:
Many thanks!

Issue #230, opened 2020-04-08
Support batched JSONRPC requests


romanz:
Following #228

vlad20012:
What do you think about using https://github.com/paritytech/jsonrpc/tree/master/tcp ?

romanz:
Thanks for the suggestion, will take a look :)

vesparny:
Hi, is this feature being considered for future implementation?
Batched requests would definitely give a performance boost, especially for wallets containing multiple transactions

Kixunil:
@vesparny if you're asking because you're experiencing performance issues it could be actually caused by inefficiency in some `electrs` code. We were brainstorming about it in #308 

If you have the time, you could check the `next` branch (and accompanied bitcoind fork) to see if your issue is fixed there. The branch still has to be changed but at least we would have more data.

craigraw:
I believe that when Electrs is used over Tor (which is increasingly becoming the default), there is no improvement that will have a bigger impact on perceived performance. To illustrate this I created a custom build of Sparrow, loaded the same wallet twice and configured a connection to an ElectrumX server over Tor. One of the wallets used batched JSONRPC, the other did not. The difference was stark - the batched load completed in 7 secs while the non-batched in 68 seconds. Here are the loading logs - same wallet, same server:

![batching](https://user-images.githubusercontent.com/862166/118971979-55175400-b970-11eb-9071-baddd106d428.gif)

Kixunil:
@craigraw I believe you're performing requests one-by-one, each issued only after previous completes, right? In that case it should be possible to send several requests in parallel to achieve better performance.

Of course I do agree that batching is just better. If parallel requests are easy to implement for you it might be a useful temporary solution.

romanz:
Please take a look at #405 :)

craigraw:
For the purposes of detecting and switching to batched JSON-RPC, can the 0.9.0 version be used (from the `client_name` of the `server.version` request)?

romanz:
It's possible, but I would suggest sending a simple "batched" request to see if the server actually supports that :)
For example (note the `[...]` brackets):
```
$ echo '[{"jsonrpc": "2.0", "method": "server.ping", "id": 0}]' | netcat 127.0.0.1 50001
[{"id":0,"jsonrpc":"2.0","result":null}]
```

Overtorment:
to my experience, you might accidentally crash an electrum server if you send unsupported request 😂

romanz:
[...]

PR #405, opened 2021-05-21
Add initial support for JSON-RPC batching

romanz:
Following #230 - please note that this PR is based over `p2p` branch (as described [here](https://github.com/romanz/electrs/issues/308#issuecomment-806061229)).

### DIFF
--- a/contrib/client.py
+++ b/contrib/client.py
@@ -7,13 +7,16 @@ class Client:
         self.f = self.s.makefile('r')
         self.id = 0
 
-    def call(self, method, *args):
-        req = {
+    def request(self, method, *args):
+        self.id += 1
+        return {
             'id': self.id,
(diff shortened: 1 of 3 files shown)


Kixunil (approved):
LGTM

craigraw:
I've spent a few hours testing this and I can't find any issues. It works really well!

In fact the only issue I found is I think unrelated to this change, but I did experience with this code so reporting it here. While tuning the read timeouts for batched calls the electrs server process occasionally terminated with the following error:
```
Error: server failed

Caused by:
    0: failed to send response: {"jsonrpc":"2.0","method":"blockchain.headers.subscribe","params":[{"height":1976465,"hex":"000000203f3..."}]}

    1: Broken pipe (os error 32)
```

PR #415, opened 2021-06-25
Add python example in RPC examples

moneroexamples:
Current `usage.md` contains only one example in `RPC examples` which uses `echo` and `netcat`. The PR adds new and corresponding example in `Python` which can speed up the initial use of the `electrs` for people who try to use the project for the first time.

### DIFF
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -452,13 +452,14 @@ $ echo '{"jsonrpc": "2.0", "method": "server.version", "params": ["", "1.4"], "i
 Corresponding example in `Python`:
 
 ```
+import json
 import socket
 
 with socket.socket(socket.AF_INET, socket.SOCK_STREAM) as s:
     s.connect(("127.0.0.1", 50001))
     f = s.makefile()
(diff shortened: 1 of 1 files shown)


Kixunil (commented):
I would prefer this in some common repository describing the protocol but don't mind it being here. Also maybe use `json.dumps` to get more realistic code?

moneroexamples:
I think you mean, `json.loads` as this will parse the raw response from the server? I kept the example very short, as it allows showcases in one glance how to interact with the server.

romanz (approved):
Thanks!

Kixunil:
@moneroexamples I meant `dumps` because I don't expect people will construct json manually.

moneroexamples:
@Kixunil 

Oh I see. You mean for input data, such as:

```
    message = json.dumps({"jsonrpc": "2.0", "method": "server.version", "params": ["", "1.4"], "id": "0"})
    s.sendall((message+'\n').encode())
```

Sure, no problem. I can update the code and make new PR if that would be ok?

Kixunil:
Yes, that' what I meant (but parsing also makes sense).

PR #417, opened 2021-06-29
Added json.dumps to Python example RPC examples

moneroexamples:
As suggested in comment https://github.com/romanz/electrs/pull/415#issuecomment-870345086 , `json.dumps` is added in-place of a manual construction of a request message into the Python example RPC examples


romanz:
Thanks!

PR #425, opened 2021-07-07
Inside contrib/addr.py example, add regtest and signet networks

aitorjs:
Tested in both regtest and (custom) signet .

### DIFF
--- a/contrib/addr.py
+++ b/contrib/addr.py
@@ -8,11 +8,19 @@ import client
 def main():
     parser = argparse.ArgumentParser()
     parser.add_argument('--testnet', action='store_true')
+    parser.add_argument('--regtest', action='store_true')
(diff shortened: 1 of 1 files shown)


romanz:
Thanks!

contrib/addr.py
```diff
@@ -8,12 +8,20 @@
...
 
     if args.testnet:
         port = 60001
         from pycoin.symbols.xtn import network
+    if args.regtest:
+        port = 60401
+        from pycoin.symbols.xrt import network
+    if args.signet:
+        port = 60601
+        from pycoin.symbols.xtn import network
```
Kixunil, line 24: Would be nice to explicitly check if only one is specified. Although not important since it's a simple example.

PR #435, opened 2021-07-21
Force MSRV in Travis

Kixunil:
This should hopefully prevent CI from using wrong Rust version.
Strictly speaking, rustup should use the version from `rust-toolchain` file but it looks like it was ignored.


Kixunil:
Since the test didn't fail, it's actually wrong.

Kixunil:
OMG, this uses GitHub actions and I modified travis. :man_facepalming:

PR #438, opened 2021-07-21
Set toolchain in GitHub actions to 1.41.1

Kixunil:
This should make sure to catch compatibility mistakes in commits.

I used some nicer trick in the past but at first I'd like to see if this helps at all.


romanz:
Many thanks!

PR #442, opened 2021-07-22
Fix `--auth` support

romanz:
If not specified, use `--cookie-file`.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -105,7 +105,7 @@ impl FromStr for BitcoinNetwork {
 }
 
 impl ::configure_me::parse_arg::ParseArgFromStr for BitcoinNetwork {
-    fn describe_type<W: fmt::Write>(mut writer: W) -> std::fmt::Result {
+    fn describe_type<W: fmt::Write>(mut writer: W) -> fmt::Result {
         write!(writer, "either 'bitcoin', 'testnet', 'regtest' or 'signet'")
     }
 }
@@ -123,7 +123,7 @@ pub struct Config {
     pub network: Network,
     pub db_path: PathBuf,
     pub daemon_dir: PathBuf,
-    pub daemon_auth: Auth,
+    pub daemon_auth: SensitiveAuth,
     pub daemon_rpc_addr: SocketAddr,
     pub daemon_p2p_addr: SocketAddr,
     pub electrum_rpc_addr: SocketAddr,
@@ -133,6 +133,27 @@ pub struct Config {
     pub ignore_mempool: bool,
     pub server_banner: String,
     pub args: Vec<String>,
+}
+
+pub struct SensitiveAuth(pub Auth);
+
+impl SensitiveAuth {
+    pub(crate) fn get_auth(&self) -> Auth {
+        self.0.clone()
+    }
+}
+
+impl fmt::Debug for SensitiveAuth {
+    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
+        match self.0 {
+            Auth::UserPass(ref user, _) => f
+                .debug_tuple("UserPass")
+                .field(&user)
+                .field(&"<sensitive>")
+                .finish(),
+            _ => write!(f, "{:?}", self.0),
+        }
+    }
 }
 
 /// Returns default daemon directory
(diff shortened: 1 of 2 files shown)


src/config.rs
```diff
@@ -232,11 +233,23 @@ impl Config {
             .cookie_file
             .unwrap_or_else(|| daemon_dir.join(".cookie"));
 
+        let daemon_auth = match config.auth {
+            Some(user_password) => {
+                let parts: Vec<&str> = user_password.split(":").collect();
+                if parts.len() != 2 {
```
Kixunil, line 239: I don't believe `:` in password is forbidden. I suggest:

```rust
let colon_pos = user_password.find(':').unwrap_or_else(|| { eprintln!("Error: auth cookie doesn't contain colon"); std::process::exit(1) });
Auth::UserPass(user_password[..colon_pos].to_owned(), user_password[(colon_pos + 1)..].to_owned())
```

Note that my suggestion also intentionally avoids leaking `auth`.

src/daemon.rs
```diff
@@ -48,11 +48,12 @@ fn rpc_poll(client: &mut bitcoincore_rpc::Client) -> PollResult {
 
 pub(crate) fn rpc_connect(config: &Config) -> Result<bitcoincore_rpc::Client> {
     let rpc_url = format!("http://{}", config.daemon_rpc_addr);
-    if !config.daemon_cookie_file.exists() {
-        bail!("{:?} is missing", config.daemon_cookie_file);
+    if let bitcoincore_rpc::Auth::CookieFile(ref path) = config.daemon_auth {
+        if !path.exists() {
+            bail!("{:?} is missing", path);
+        }
```
Kixunil, line 54: Isn't it better to let RPC code check this? This is TOCTOU issue and I don't see any additional benefit.

src/config.rs
```diff
@@ -232,11 +233,23 @@ impl Config {
             .cookie_file
             .unwrap_or_else(|| daemon_dir.join(".cookie"));
 
+        let daemon_auth = match config.auth {
```
Kixunil, line 236: There's no check whether both option are specified - that should be rejected as ambiguous. I usually do that by writing `match (config.cookie_file, confi.auth)` and enumerating all possibilities.

Kixunil (changes requested):
I think at least that ambiguity check should be implemented.

src/config.rs
romanz: You're right - fixed using `str::splitn`.

src/daemon.rs
romanz: Unfortunately, RPC error message is not very clear in case the cookie file is missing:
```
Error: failed to connect to RPC: 127.0.0.1:8332

Caused by:
    I/O error: No such file or directory (os error 2)
```

src/config.rs
romanz: Sounds good, fixed.

src/config.rs
```diff
@@ -122,7 +123,7 @@ pub struct Config {
     pub network: Network,
     pub db_path: PathBuf,
     pub daemon_dir: PathBuf,
-    pub daemon_cookie_file: PathBuf,
+    pub daemon_auth: Auth,
```
romanz, line 126: Make sure it's not dumped into the log.

src/daemon.rs
Kixunil: Yeah, io::Error not having path is the single most annoying thing about Rust. :D
Let's keep it this way and fix the error in a separate issue/PR.

src/config.rs
```diff
@@ -228,15 +229,28 @@ impl Config {
         }
 
         let daemon_dir = &config.daemon_dir;
-        let daemon_cookie_file = config
-            .cookie_file
-            .unwrap_or_else(|| daemon_dir.join(".cookie"));
+        let daemon_auth = match (config.auth, config.cookie_file) {
+            (None, None) => Auth::CookieFile(daemon_dir.join(".cookie")),
+            (None, Some(cookie_file)) => Auth::CookieFile(cookie_file),
+            (Some(cookie), None) => {
```
Kixunil, line 235: Nit: I'd prefer calling this `auth` instead of `cookie` - same reason we renamed the parameter. :)

Kixunil (approved):
Apart from the nit it looks good now.

src/config.rs
romanz: https://github.com/romanz/electrs/pull/442/commits/ad46aa37ca9fc7d88f55226cbf3c4c07917fe24b

src/config.rs
romanz: Good catch, thanks!

romanz: Fixed in https://github.com/romanz/electrs/pull/442/commits/8ef15e25b9d24e16626a6157b07de9d716162dce

Kixunil (approved):
Awesome!

PR #445, opened 2021-07-23
Add server.features RPC

### LINKED ISSUE #439: server.features protocol method not implemented
Jul 22 11:02:06 aaa electrs[25439]: 2021-07-22T11:02:06.326+01:00 - TRACE - RPC Request("{\"jsonrpc\":\"2.0\",\"method\":\"server.features\",\"params\":[],\"id\":6}\n")
Jul 22 11:02:06 aaa electrs[25439]: 2021-07-22T11:02:06.326+01:00 - ERROR - [127.0.0.1:41596] connection handling failed: Error: unknown method server.features []

ver 0.8.9

romanz:
Fixes #439

### DIFF
--- a/src/electrum.rs
+++ b/src/electrum.rs
@@ -116,6 +116,7 @@ pub struct Rpc {
     rpc_duration: Histogram,
     daemon: Daemon,
     banner: String,
+    port: u16,
 }
 
 impl Rpc {
@@ -130,6 +131,7 @@ impl Rpc {
             rpc_duration,
             daemon: Daemon::connect(config)?,
             banner: config.server_banner.clone(),
+            port: config.electrum_rpc_addr.port(),
         })
     }
 
@@ -327,11 +329,14 @@ impl Rpc {
         Ok(json!(self.tracker.fees_histogram()))
     }
 
+    fn server_id(&self) -> String {
+        format!("electrs/{}", ELECTRS_VERSION)
+    }
+
     fn version(&self, (client_id, client_version): (String, Version)) -> Result<Value> {
         match client_version {
             Version::Single(v) if v == PROTOCOL_VERSION => {
-                let server_id = format!("electrs/{}", ELECTRS_VERSION);
-                Ok(json!([server_id, PROTOCOL_VERSION]))
+                Ok(json!([self.server_id(), PROTOCOL_VERSION]))
             }
             _ => {
                 bail!(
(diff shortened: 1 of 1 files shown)


romanz:
CC: Overtorment

Overtorment (approved):

src/electrum.rs
```diff
@@ -344,6 +344,18 @@ impl Rpc {
         }
     }
 
+    fn features(&self) -> Result<Value> {
+        Ok(json!({
+            "genesis_hash": self.tracker.chain().get_block_hash(0),
+            "hosts": {},
```
Kixunil, line 350: Should we add `"localhost": { "tcp_port": electrum_rpc_port }`?

src/electrum.rs
```diff
@@ -344,6 +344,18 @@ impl Rpc {
...
+            "protocol_max": PROTOCOL_VERSION,
+            "protocol_min": PROTOCOL_VERSION,
+            "pruning": null,
+            "server_version": format!("electrs {}", ELECTRS_VERSION),
```
Kixunil, line 354: The (recent) doc says:

> Should be the same as the first element of the result to the server.version() RPC call.

However here's it's separated with space but [above](https://github.com/romanz/electrs/pull/445/files#diff-ede2fa6030c1edb89d5de41ac20473c31669e46197974a46108a9bff88eebfaeR333) with slash.
I suggest writing `fn server_id() -> String` and using it in both places to keep them in sync

Kixunil (commented):
Looks good apart from small detail

romanz:
Tested locally with:
```
$ echo '{"jsonrpc":"2.0","method":"server.features","id":123}' | netcat localhost 50001 | jq .
{
  "id": 123,
  "jsonrpc": "2.0",
  "result": {
    "genesis_hash": "000000000019d6689c085ae165831e934ff763ae46a2a6c172b3f1b60a8ce26f",
    "hash_function": "sha256",
    "hosts": {
      "tcp_port": 50001
    },
    "protocol_max": "1.4",
    "protocol_min": "1.4",
    "pruning": null,
    "server_version": "electrs/0.9.0"
  }
}
```

src/electrum.rs
romanz: Good catch - thanks!

src/electrum.rs
romanz: :+1:

PR #449, opened 2021-07-29
Add a workaround for latest Bitcoin Core

romanz:
See https://github.com/rust-bitcoin/rust-bitcoincore-rpc/pull/190.

### DIFF
--- a/src/daemon.rs
+++ b/src/daemon.rs
@@ -17,7 +17,7 @@ enum PollResult {
 }
 
 fn rpc_poll(client: &mut bitcoincore_rpc::Client) -> PollResult {
-    match client.get_blockchain_info() {
+    match client.call::<BlockchainInfo>("getblockchaininfo", &[]) {
         Ok(info) => {
             let left_blocks = info.headers - info.blocks;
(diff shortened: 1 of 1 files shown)


Kixunil:
Does this incompatibility affect master? (will review more properly later)

romanz:
I think it should not:
https://github.com/romanz/electrs/blob/3ef38fc6f1eefa924f1d67273d383f0e5578374a/src/daemon.rs#L98-L106

src/daemon.rs
```diff
@@ -82,8 +96,8 @@ impl Daemon {
         if !network_info.network_active {
             bail!("electrs requires active bitcoind p2p network");
         }
-        let blockchain_info = rpc.get_blockchain_info()?;
-        if blockchain_info.pruned {
+        let info: BlockchainInfo = rpc.call("getblockchaininfo", &[])?;
+        if info.pruned {
```
Kixunil, line 100: Nit: keeping the same name would make diff easier to understand.

Kixunil (approved):
Looks good

Issue #451, opened 2021-07-31
Speeding up getrawtransaction when txindex=1


Pantamis:
The txindex of Core stores the exact location of a transaction in the blockfiles. This allows to retrieve the full transaction from its transaction ID in one short disk read with `getrawtransaction`. This RPC call also allows to search for a transaction in a block by specifying the blockhash of where it should be as argument, txindex enable or not. That's how electrs is able to work without txindex: it stores the blockheight of any transaction in its index (and the txid is the key).

But there is an issue: if the blockhash is given, Core will read the whole block and search the transaction inside, no mater if txindex is enable or not !!

This means that giving the blockhash to `getrawtransaction` in fact slower transaction retrieval if txndex is enable ! Indeed, the txindex is used only if the transaction wasn't found in the block so adding the blockhash will result in at least one full block disk read (much longer !). This was pointed in bitcoin/bitcoin#22382 and recently solved by bitcoin/bitcoin#22383, merged two weeks ago, well after feature freeze.

This means the issue is still present in Core 22.0 release and will ultimately be solved in 23.0 release. In the meantime, each time Electrs retrieves a ransaction in a block, it suffers from lower perf because it always fills the blockhash argument of `getrawtransaction` so Core doesn't use `txindex`, this increases time to retrieves a transaction by a factor of 5 in average.

It could be nice to detect if `txindex=1` in Core or add a config flag so that when set ElectRS always call `getrawtransaction` without adding the blockhash, to benefit from the higher perfs with txindex. This would be extremly beneficial to current branch. (less for p2p branch because it indexes blockheight, so the whole block are always read no matter what) Once 23.0 is out, this flag can be removed.

The other solution is to implement that in rust-bitcoincore-rpc instead...

Kixunil:
> detect if `txindex=1` in Core or add a config flag

Please detect - much more reliable than people setting flags. I'd also wait with removing it to not create dependency hell.

Pantamis:
The RPC command `getindexinfo "txindex"` returns nothing if txindex=0 and 
```
[
    {
        "txindex": {
            "synced": true,
            "best_block_height": ...
        }
    }
]
``` 
if `txindex=1`.

But this call is not supported in rust-bitcoincore-rpc lib. Maybe I should first open a PR there first ?

Kixunil:
[...]

PR #452, opened 2021-08-02
Check txindex and use blockhash if disabled

Pantamis:
Solve #451 

Use value in synced field for now.

### DIFF
--- a/src/daemon.rs
+++ b/src/daemon.rs
@@ -113,11 +113,27 @@ struct NetworkInfo {
 }
 
 #[derive(Serialize, Deserialize, Debug)]
-struct IndexInfo {
+struct TxIndexInfo {
     synced: bool,
     best_block_height: usize,
 }
 
+#[derive(Deserialize, Debug)]
+#[serde(rename_all = "lowercase")]
+enum IndexInfo {
+    TxIndex(TxIndexInfo),
+}
+
+enum TxIndexState {
+    /// Didn't check yet, re-check next time
+    Unknown,
+    /// assume unavailable but re-check after a minute since `last_check`
+    Syncing { last_check: std::time::Instant, },
+    /// Available and synced - do not pass blockhash
+    Synced,
+    /// `txindex=0` not configured - pass blockhash and don't retry check
+    Unavailable,
+}
 
 pub struct MempoolEntry {
     fee: u64,   // in satoshis
@@ -465,9 +481,7 @@ impl Daemon {
     }
 
     fn getindexinfo(&self, indexes: Vec<&str>) -> Result<Vec<IndexInfo>> {
-        let indexes: Vec<Value> = indexes.iter().map(|index| json!([index])).collect();
-        let info: Value = self.request("getindexinfo", json!(indexes))?;
-        from_value(info).chain_err(|| "invalid index info")
+        from_value(self.request("getindexinfo", json!(indexes))?).chain_err(|| "invalid index info")
     }
 
     pub fn get_subversion(&self) -> Result<String> {
@@ -534,7 +548,19 @@ impl Daemon {
         blockhash: Option<BlockHash>,
     ) -> Result<Transaction> {
         let mut args = json!([txhash.to_hex(), /*verbose=*/ false]);
-        if let (true, Some(blockhash)) = (self.getindexinfo(vec!["txindex"])?[0].synced, blockhash) {
+        let txindex_available = self.getindexinfo(vec!["txindex"])
+            .unwrap_or_else(|error| {
+                error!("failed to get index info: {}, assuming txindex unavailable", error);
+                Vec::new()
+            })
+            .iter().find_map(|info| {
+                match info {
+                    IndexInfo::TxIndex(txindex) => Some(txindex),
+                }
+            })
+            .map(|txindex| txindex.synced)
+            .unwrap_or(false);
+        if let (false, Some(blockhash)) = (txindex_available, blockhash) {
             args.as_array_mut().unwrap().push(json!(blockhash.to_hex()));
         }
         tx_from_value(self.request("getrawtransaction", args)?)
@@ -547,7 +573,19 @@ impl Daemon {
         verbose: bool,
     ) -> Result<Value> {
         let mut args = json!([txhash.to_hex(), verbose]);
-        if let (true, Some(blockhash)) = (self.getindexinfo(vec!["txindex"])?[0].synced, blockhash) {
+        let txindex_available = self.getindexinfo(vec!["txindex"])
+            .unwrap_or_else(|error| {
+                error!("failed to get index info: {}, assuming txindex unavailable", error);
+                Vec::new()
+            })
+            .iter().find_map(|info| {
+                match info {
+                    IndexInfo::TxIndex(txindex) => Some(txindex),
+                }
+            })
+            .map(|txindex| txindex.synced)
+            .unwrap_or(false);
+        if let (false, Some(blockhash)) = (txindex_available, blockhash) {
             args.as_array_mut().unwrap().push(json!(blockhash.to_hex()));
         }
         self.request("getrawtransaction", args)
--- a/src/daemon.rs
+++ b/src/daemon.rs
@@ -6,7 +6,7 @@ use bitcoin::hashes::hex::{FromHex, ToHex};
 use bitcoin::hashes::Hash;
 use bitcoin::network::constants::Network;
 use serde_json::{from_str, from_value, Map, Value};
-use std::{collections::{HashMap, HashSet}, time::Instant};
+use std::collections::{HashMap, HashSet};
 use std::io::{BufRead, BufReader, Lines, Write};
 use std::net::{SocketAddr, TcpStream};
 use std::path::PathBuf;
@@ -174,6 +174,7 @@ struct Connection {
     cookie_getter: Arc<dyn CookieGetter>,
     addr: SocketAddr,
     signal: Waiter,
+    txindex: TxIndexState,
 }
 
 fn tcp_connect(addr: SocketAddr, signal: &Waiter) -> Result<TcpStream> {
@@ -206,6 +207,7 @@ impl Connection {
             cookie_getter,
             addr,
             signal,
+            txindex: TxIndexState::Unknown,
         })
     }
 
@@ -287,6 +289,24 @@ impl Connection {
             );
         })
     }
+
+    fn update_txindex_state(&mut self) -> Result<()> {
+        let request = json!({"method": "getindexinfo", "params": "txindex", "id": 0}).to_string();
+        self.send(&request)?;
+        let response = self.recv()?;
+        let result_vec = from_str::<Vec<TxIndexInfo>>(&response).chain_err(|| "invalid JSON")?;
+        match result_vec.get(0) {
+            Some(txindex) => {
+                if txindex.synced {
+                    self.txindex = TxIndexState::Synced;
+                } else {
+                    self.txindex = TxIndexState::Syncing{ last_check: Instant::now() };
+                }
+            }
+            None => { self.txindex = TxIndexState::Unavailable; }
+        }
+        Ok(())
+    }
 }
 
 struct Counter {
@@ -316,9 +336,6 @@ pub struct Daemon {
     // monitoring
     latency: HistogramVec,
     size: HistogramVec,
-
-    // index available
-    txindex: TxIndexState,
 }
 
 impl Daemon {
@@ -353,7 +370,6 @@ impl Daemon {
                 HistogramOpts::new("electrs_daemon_bytes", "Bitcoind RPC size (in bytes)"),
                 &["method", "dir"],
             ),
-            txindex: TxIndexState::Unknown,
         };
         let network_info = daemon.getnetworkinfo()?;
         info!("{:?}", network_info);
@@ -396,7 +412,6 @@ impl Daemon {
             blocktxids_cache: Arc::clone(&self.blocktxids_cache),
             latency: self.latency.clone(),
             size: self.size.clone(),
-            txindex: TxIndexState::Unknown,
         })
     }
 
@@ -490,23 +505,6 @@ impl Daemon {
         from_value(self.request("getindexinfo", json!(indexes))?).chain_err(|| "invalid index info")
     }
 
-    fn update_txindex_state(&mut self) {
-        match getindexinfo(self, vec!["txindex"]).iter().next() {
-            Some(TxIndex(txindex)) => {
-                if txindex.synced {
-                    self.txindex = TxIndexState::Synced;
-                } else {
-                    self.txindex = TxIndexState::Syncing{ last_check: Instant::now() };
-                }
-
-            }
-            None => { self.txindex = TxIndexState::Unavailable; }
-            Some(_) => { self.txindex = TxIndexState::Unknown;
-                bail!("Unexpected return from getindexinfo")
-            }
-        }
-    }
-
     pub fn get_subversion(&self) -> Result<String> {
         Ok(self.getnetworkinfo()?.subversion)
     }
@@ -571,19 +569,20 @@ impl Daemon {
         blockhash: Option<BlockHash>,
     ) -> Result<Transaction> {
         let mut args = json!([txhash.to_hex(), /*verbose=*/ false]);
-        match (self.txindex, blockhash) {
+        let mut conn = self.conn.lock().unwrap();
+        match (conn.txindex, blockhash) {
             (TxIndexState::Synced, _) => {}
             (_, None) => {}
             (TxIndexState::Unavailable, Some(blockhash)) => {
                 args.as_array_mut().unwrap().push(json!(blockhash.to_hex()));
             }
             (TxIndexState::Unknown, Some(blockhash)) => {
-                self.update_txindex_state(); 
+                conn.update_txindex_state()?; 
                 args.as_array_mut().unwrap().push(json!(blockhash.to_hex())); 
             }
-            (TxIndexState::Syncing(last_check), Some(blockhash)) => {
+            (TxIndexState::Syncing{last_check}, Some(blockhash)) => {
                 if last_check.elapsed() > Duration::from_secs(3600) {
-                    self.update_txindex_state(); 
+                    conn.update_txindex_state()?; 
                 }
                 args.as_array_mut().unwrap().push(json!(blockhash.to_hex())); 
             }
@@ -598,19 +597,20 @@ impl Daemon {
         verbose: bool,
     ) -> Result<Value> {
         let mut args = json!([txhash.to_hex(), verbose]);
-        match (self.txindex, blockhash) {
+        let mut conn = self.conn.lock().unwrap();
+        match (conn.txindex, blockhash) {
             (TxIndexState::Synced, _) => {}
             (_, None) => {}
             (TxIndexState::Unavailable, Some(blockhash)) => {
                 args.as_array_mut().unwrap().push(json!(blockhash.to_hex()));
             }
             (TxIndexState::Unknown, Some(blockhash)) => {
-                self.update_txindex_state(); 
+                conn.update_txindex_state()?; 
                 args.as_array_mut().unwrap().push(json!(blockhash.to_hex())); 
             }
-            (TxIndexState::Syncing(last_check), Some(blockhash)) => {
+            (TxIndexState::Syncing{last_check}, Some(blockhash)) => {
                 if last_check.elapsed() > Duration::from_secs(3600) {
-                    self.update_txindex_state(); 
+                    conn.update_txindex_state()?; 
                 }
                 args.as_array_mut().unwrap().push(json!(blockhash.to_hex())); 
             }
--- a/src/daemon.rs
+++ b/src/daemon.rs
@@ -6,13 +6,13 @@ use bitcoin::hashes::hex::{FromHex, ToHex};
 use bitcoin::hashes::Hash;
 use bitcoin::network::constants::Network;
 use serde_json::{from_str, from_value, Map, Value};
-use std::collections::{HashMap, HashSet};
+use std::{collections::{HashMap, HashSet}, time::Instant};
 use std::io::{BufRead, BufReader, Lines, Write};
 use std::net::{SocketAddr, TcpStream};
 use std::path::PathBuf;
 use std::sync::atomic::{AtomicU64, Ordering};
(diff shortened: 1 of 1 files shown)


src/daemon.rs
```diff
@@ -112,6 +112,13 @@ struct NetworkInfo {
     relayfee: f64, // in BTC
 }
 
+#[derive(Serialize, Deserialize, Debug)]
+struct IndexInfo {
+    synced: bool,
+    best_block_height: usize,
+}
```
Kixunil, line 119: I suggest renaming this to `TxIndexInfo`. (or `TxindexInfo`? the former reads better IMO, the latter seems formally consistent)

Then have a helper enum:

```rust
[derive(Deserialize, Debug)]
#[serde(rename_all = "lowercase")]
enum IndexInfo {
    TxIndex(TxIndexInfo),
}
```

Then try to deserialize `Vec<IndexInfo>` directly from the response:

```rust
fn getindexinfo(&self, indexes: Vec<&str>) -> Result<Vec<IndexInfo>> {
    self.request("getindexinfo", json!(indexes)).map_err(Into::into)
}
```

**Why enum instead of struct?**

This is preparation for the possibility of adding other index infos.
If this is upstreamed to `bitcoincore-rpc` it'd be the best way wit additional `#[non_exhaustive]` attribute (but not sure if its MSRV supports it OTOH maybe it can be bumped). It will force us to add `_ => None` match arm below, but that's fine.

I have an alternative black magic idea that would statically guarantee specific type (for our use case) but it's not exclusive because I think the file argument is optional.

src/daemon.rs
```diff
@@ -514,13 +527,14 @@ impl Daemon {
...
 
+
     pub fn gettransaction(
         &self,
         txhash: &Txid,
         blockhash: Option<BlockHash>,
     ) -> Result<Transaction> {
         let mut args = json!([txhash.to_hex(), /*verbose=*/ false]);
-        if let Some(blockhash) = blockhash {
+        if let (true, Some(blockhash)) = (self.getindexinfo(vec!["txindex"])?[0].synced, blockhash) {
```
Kixunil, line 537: I think the logic is inverted here: it should only add the blockhash if it's **not** synced.
There's also one more problem: according to what you said: "RPC command `getindexinfo "txindex"` returns nothing if txindex=0" this will panic if `txindex=0` because you access it unconditionally.

Also it'd be nice to **not** blow up the whole thing if `indexinfo` can't be obtained - just log it and keep degraded performance.

I suggest:

```rust
let txindex_available = self.getindexinfo(vec!["txindex"])
    .unwrap_or_else(|error| {
        error!("failed to get index info: {}, assuming txindex unavailable", error));
        Vec::new()
    })
    .find_map(|info| {
        match info {
            IndexInfo::TxIndex(txindex) => Some(txindex),
        }
    })
    .map(|txindex| txindex.synced)
    .unwrap_or(false);

    if let (false, Some(blockhash) = (txindex_available, blokhash) {
    // ...
```

Note that `synced` could be mapped directly in `find_map` but I find this clearer. Do you agree @romanz ?

src/daemon.rs
```diff
@@ -533,7 +547,7 @@ impl Daemon {
         verbose: bool,
     ) -> Result<Value> {
         let mut args = json!([txhash.to_hex(), verbose]);
-        if let Some(blockhash) = blockhash {
+        if let (true, Some(blockhash)) = (self.getindexinfo(vec!["txindex"])?[0].synced, blockhash) {
```
Kixunil, line 550: Same as above

Kixunil (changes requested):
Thanks!

There's a crash issue that needs to be fixed - see comments.

Additionally to what I wrote below, I think we should cache the result of `getindexinfo` somehow - otherwise the performance may be actually worse. However it's not as easy as it may look. We need to somehow invalidate the state when connection drops and also need to occasionally re-check if txindex synced.

I would store the state as:

```rust
enum TxIndexState {
    /// Didn't check yet, re-check next time
    Unknown,
    /// assume unavailable but re-check after a minute since `last_check`
    Syncing { last_check: std::time::Instant, },
    /// Available and synced - do not pass blockhash
    Synced,
    /// `txindex=0` not configured - pass blockhash and don't retry check
    Unavailable,
}
```

Then have a helper method that will update the state if needed and return `bool` and have some handler for dropped connection to reset the state to `Unknown`.

src/daemon.rs
Pantamis: I really like the enum I am not familiar with it but it really looks much better indeed fo futur index types :)

src/daemon.rs
Pantamis: Aw I made the stupid mistake x).

Yeah indeed I forgot we need to handle the case where the RPC returns nothing, my bad

Kixunil: No problem! :)

src/daemon.rs
Pantamis: I have trouble with `getindexinfo` because `self.request` return me a `Value` type.

So I deserialize it by returning `from_value(self.request("getindexinfo", json!(indexes))?).chain_err(|| "invalid index info")` is that ok or do you think there is a better way to do it ?

Kixunil: It's OK stopgap solution but long term it should be changed to deserialize into the struct directly. This will be probably cleaner in the `bitcoincore-rpc` crate, so let's wait for it with that for now and only try to solve it later.

Although that method should be changed anyway because whatever is using it may be underperformant. Should be a different PR, I believe.

Pantamis: > long term it should be changed to deserialize into the struct directly.

Do you mean it requires to change `self.request` ? Any way as you said, if it makes it to `bitcoincore-rpc` then code will looks better, no doubt.

> Although that method should be changed anyway because whatever is using it may be underperformant. Should be a different PR, I believe.

I am not sure if you are talking about using the `TxIndexState` but I am thinking about how to do it now, it is clear that asking for `txindex` value each time we query a transaction is not good :/

Kixunil: > Do you mean it requires to change `self.request` ?

Yes something along the lines: `fn request<T: serde::Deserialize>(...) -> Result<T>`

> I am not sure if you are talking about using the `TxIndexState`

No, I meant changing `request()`. Anyway this is `master` branch and I'm not sure if there will be a release before `p2p` gets finished.

> it is clear that asking for `txindex` value each time we query a transaction is not good

Yep, I'm just unsure how to handle failed connection without a huge code rewrite and forgetting something. One thing that comes to mind is wrapping the connection in a helper struct that will contain that cache and any other cache in the future. The struct is initialized with empty values after connecting.

src/daemon.rs
```diff
@@ -113,11 +113,27 @@ struct NetworkInfo {
...
+enum TxIndexState {
+    /// Didn't check yet, re-check next time
+    Unknown,
+    /// assume unavailable but re-check after a minute since `last_check`
+    Syncing { last_check: std::time::Instant, },
+    /// Available and synced - do not pass blockhash
+    Synced,
+    /// `txindex=0` not configured - pass blockhash and don't retry check
+    Unavailable,
+}
```
Kixunil, line 136: This is partial change, right?

Pantamis: Yeah, still thinking about how using it now, quite difficult as you already told x)

Kixunil: Just realized I forgot about `#[derive(Copy, Clone)]` it will not solve the problem but avoid annoyance later. :)

romanz:
Sorry for the delay, hopefully will more available in the next week :(

Pantamis:
In the last commit I put the TxIndexState in the Connection struct because I need a mutable state and the Daemon struct is not supposed to be mutable. This is weird but it should work. Maybe I didn't made all necessary checks with the RPC call to `getindexinfo`, tell me if it is an issue.

The TxIndexState is Unknown when connection is initialised. It is checked when the daemon must retrieve transactions and only if it is Unknown or was Syncing one hour ago. The daemon do not add the blockhash to the `getrawtransaction` if the TxIndexState is Synced, which should increase perf.

@romanz It's ok, @Kixunil already made this PR much much better :) it should be easier to review now (I hope so !)

src/daemon.rs
```diff
@@ -533,8 +597,23 @@ impl Daemon {
         verbose: bool,
     ) -> Result<Value> {
         let mut args = json!([txhash.to_hex(), verbose]);
-        if let Some(blockhash) = blockhash {
-            args.as_array_mut().unwrap().push(json!(blockhash.to_hex()));
+        let mut conn = self.conn.lock().unwrap();
```
Kixunil, line 600: Making RPC calls under lock makes my hair stand up. The risk of deadlocks is too great IMO and we already had deadlock issues so I'd really, really like to avoid it. (Disclaimer: I think we already have such things and I think they are not good and there's no good reason to repeat such risky design.)

src/daemon.rs
```diff
@@ -533,8 +597,23 @@ impl Daemon {
...
+            (_, None) => {}
+            (TxIndexState::Unavailable, Some(blockhash)) => {
+                args.as_array_mut().unwrap().push(json!(blockhash.to_hex()));
+            }
+            (TxIndexState::Unknown, Some(blockhash)) => {
+                conn.update_txindex_state()?; 
+                args.as_array_mut().unwrap().push(json!(blockhash.to_hex())); 
+            }
+            (TxIndexState::Syncing{last_check}, Some(blockhash)) => {
+                if last_check.elapsed() > Duration::from_secs(3600) {
```
Kixunil, line 612: 1h seems quite pessimistic to me. AFAIR syncing is not terribly slow so I'd prefer trading a little bit of performance during sync for better performance after it. 10 mins maybe?

src/daemon.rs
```diff
@@ -112,6 +112,30 @@ struct NetworkInfo {
...
+#[serde(rename_all = "lowercase")]
+enum IndexInfo {
+    TxIndex(TxIndexInfo),
+}
+
+#[derive(Copy, Clone)]
+enum TxIndexState {
+    /// Didn't check yet, re-check next time
+    Unknown,
+    /// assume unavailable but re-check after a minute since `last_check`
```
Kixunil, line 131: This comment also says after a minute while the code waits an hour. They should be the same once we agree on the period.

src/daemon.rs
Pantamis: I think I was worried about the cost of calling `getindexinfo` for each transaction (the reason why we introduced TxIndexState at the first place). But 10 min seems fair too, this call is quite fast after all :)

src/daemon.rs
Pantamis: Gonna make it 10 minutes !

src/daemon.rs
Pantamis: I don't know if there is a simple way to avoid that here. I think this has to do with the fact that we shouldn't do that here at the first place (but Core should have).

The other solution I see is to add txindex as field of Daemon struct and use mutable references to Daemon each time we need to get a transaction. I find it quite disruptive...

Kixunil:
> the Daemon struct is not supposed to be mutable

I think you mean it's shared? You can mutate shared structs if you do it via safe method such as mutex, rwlock or atomic variables but `Connection` is actually a good place here because it will get cleared automatically when the connection fails.

The best way to do it **in p2p** I can think of now is having the state inside `Mutex` and have another state `Checking(CondVar)` where threads register to be woken up by the checker thread after the check is done so that multiple threads don't attempt to check same thing simultaneously. Obviously one would lock the mutex, check current state and if the state needs updating register condvar and unlock. This should be combined with helper atomic to avoid Mutex in case the state doesn't need checking anymore.

Pantamis:
> > the Daemon struct is not supposed to be mutable
> 
> I think you mean it's shared? You can mutate shared structs if you do it via safe method such as mutex, rwlock or atomic variables but `Connection` is actually a good place here because it will get cleared automatically when the connection fails.
> 
> The best way to do it **in p2p** I can think of now is having the state inside `Mutex` and have another state `Checking(CondVar)` where threads register to be woken up by the checker thread after the check is done so that multiple threads don't attempt to check same thing simultaneously. Obviously one would lock the mutex, check current state and if the state needs updating register condvar and unlock. This should be combined with helper atomic to avoid Mutex in case the state doesn't need checking anymore.

Really interesting, I have so much to learn. Indeed you are right, it seems more appropriate to hold IndexState in Connection struct.

As I said in an other comment, the perf improvement is really worth it in `master` but not so much in `p2p` because `p2p` is not using `txid` anywhere except when you already know the `txid`. So we would only see perf improvments when the user must retrieve a transaction from its `txid`, there is no improvment possible when retrieving address history or spending transaction in `p2p` because we must already load the whole block that contains it to find it.

In `master`, [`load_txn`](https://github.com/romanz/electrs/blob/7ef39f4d467e7620c998ea95a2902cd89fc30559/src/query.rs#L406) is used to scan the set of candidate transactions and it used its knowledge of the `txid` we get from the store index. So the perf would greatly benefit from `txindex=1`

Kixunil:
> because we must already load the whole block that contains it to find it.

Ah, I missed this little detail. I think we should measure if slowness of RPC is so big that it's worth loading whole block from P2P and searching and consider using RPC if not.

Pantamis:
> > because we must already load the whole block that contains it to find it.
> 
> Ah, I missed this little detail. I think we should measure if slowness of RPC is so big that it's worth loading whole block from P2P and searching and consider using RPC if not.

There are several tradeoffs here. The `p2p` branch's index is very very small because we only store the blockheight but this requires retrieving full block when searching for a transaction under certain criteria (funding, spending or txid).

I think we can safely say if `txindex` is enabled we should discard `txid` store index, this would reduce electrs' index even more and retrieving transactions with `txindex` in Core with RPC is very very efficient ! (it directly reads the full transaction, `txindex` return the position in blockfile of the transaction with the given `txid`, ElectrumX requires `txindex` and is faster than ElectRS for example)

This however cannot be used to retrieve transactions based on scripthash and spending outpoint because we only index the blockheight, `txindex` will not help in perf here because we must retrieve the full block for each transaction anyway (and this is quite efficient with P2P). The only way to increase perf with `txindex` would be that the index return the `txid` but it implies it is bigger so...

Kixunil:
Damn, it feels like my knowledge of electrs is diminishing over time. :( You're right.

It got me thinking `p2p` stores 32 bits of block height which means it can run for 81808 years assuming 210000 blocks every four years. Block size is capped at 4M, which needs 22 bits to represent any offset. If we added 2B to every input and output we could represent block height for up to 1278 years and any offset. This is an increase in size by 11.11-16.67 % traded for possibly significant speed improvement.

Pantamis:
> Damn, it feels like my knowledge of electrs is diminishing over time. :( You're right.

I would rather say that the fact that `p2p` branch and `master` are so different and coexist since half a year explain the difficulties x)

> It got me thinking `p2p` stores 32 bits of block height which means it can run for 81808 years assuming 210000 blocks every four years. Block size is capped at 4M, which needs 22 bits to represent any offset. If we added 2B to every input and output we could represent block height for up to 1278 years and any offset. This is an increase in size by 11.11-16.67 % traded for possibly significant speed improvement.

It could be interesting to look at [what `txindex` is storing exactly](https://bitcoin.stackexchange.com/questions/28168/what-are-the-keys-used-in-the-blockchain-leveldb-ie-what-are-the-keyvalue-pair). We may also need to have the block offset in the blockfile too, which exists in block index of Core but if I understand the discussion in #308 well there is no RPC command to ask it.

What do you mean by adding "2B to every input and output" ? What I understand from your proposal: instead of using the 32 bits to store the blockheght, we use 22 bits to store an offset in the block and the 10 other for height.

Kixunil:
Electrs doesn't access block files in `p2p` directly anymore, so offset is not needed.

> What do you mean by adding "2B to every input and output" ? What I understand from your proposal: instead of using the 32 bits to store the blockheght, we use 22 bits to store an offset in the block and the 10 other for height.

4B currently + 2 additional bytes == 6B == 48 bits
That's 22 bits for offset (up to 4 MiB) and 26 bits for block height (~1278 years)

Pantamis:
> Electrs doesn't access block files in `p2p` directly anymore, so offset is not needed.

Aww I think I got it ! You want to get the block through P2P then using the stored offset to not have to scan all the transactions to find which one is spending an outpoint or using the script with the hash we are searching for.

Actually I really like this idea, I think it can improve a lot the perf at low cost !

> 4B currently + 2 additional bytes == 6B == 48 bits
> That's 22 bits for offset (up to 4 MiB) and 26 bits for block height (~1278 years)

Ok, if RocksDB is still performing well with 48 bits instead of 32 bits in stored value, it is clearly fine !

> Note that meanwhile I looked at the code and noticed whole block is parsed into transactions, allocating each - the proposed change assumes this gets optimized.

It doesn't seems hard to receive the raw block data and only start processing it when the offset is reach until we get a valid transaction. The biggest part seems too me detecting when the transaction is actually complete soon enough. Then we parse it and check that it is indeed satisfying the criteria we were searching for (scripthash, spending an outpoint or txid)

Kixunil:
> if RocksDB is still performing well with 48 bits instead of 32 bits in stored value

AFAIU RocksDB works on raw byte arrays so it shouldn't care about number of bytes.

> It doesn't seems hard to receive the raw block data and only start processing it when the offset is reach

We need to implement it ourselves though - `bitcoin` library is currently hostile to this kind of optimization. Was thinking of proposing alternative APIs some time ago but never got the time. Also would be better to have some measurement before we claim it actually improves anything. But anyway, block parsing is actually quite easy - I did it in the past to learn more about Bitcoin although without SegWit.

Pantamis:
> But anyway, block parsing is actually quite easy - I did it in the past to learn more about Bitcoin although without SegWit.

One other thing we can do then (if block parsing is quite fast) is to store the index position of the transaction in the block along with blockheight instead. The smallest possible transaction size [seems to be 61 bits](https://bitcoin.stackexchange.com/questions/54679/what-are-the-minimum-sizes-for-transactions-and-blocks-in-bitcoin). This means a block has at most 4M/(4*61) = 16394 transactions which means 14 bits are sufficient (in theory 15 but I am highly conservative already on maximum number of transaction in a block already) to get the position of the transaction.

So we can have 14 bits for position in block and 48 - 14 = 34 bits for blockheight.

This avoids having to filter every transaction of the block which already allow a good speed up if it is easy to extract the transaction from its position in the block !

Kixunil:
> which already allow a good speed up

Did you actually measure it? As far as I understand you only save the cost of hashing and comparing hashes, which I don't expect to be very high. Also making sure `electrs` can run for longer than a few hundred years sounds like waste of time. There will be new updates at least to resolve the hardfork required to fix time overflow.

However, with your scheme we could just add a *single byte* which would be 40 bits - 26 for height (same number, yay!), 14 for index. So only 5.55-8.34% increase.

Just checked that 14 vs 15 bits is just 10 tx difference. To be on the safe side we could store overflow as highest number and continue linear search if tx is not found there.

Pantamis:
> Did you actually measure it? As far as I understand you only save the cost of hashing and comparing hashes, which I don't expect to be very high. 

In fact you can save a lot more ! This is how we find a transaction with an output with a certain scripthash in blocks that the store index return in `p2p` branch for this scripthash: https://github.com/romanz/electrs/blob/94efd3a75dd72fe43fa42b75bd9e93ae41e014d5/src/status.rs#L261-L264
https://github.com/romanz/electrs/blob/94efd3a75dd72fe43fa42b75bd9e93ae41e014d5/src/status.rs#L395-L402

We iterate on all outputs of all transactions of all blocks returned by the index to find which transactions have an output with the same scripthash.

If we had the position in block of the transactions at the first place we wouldn't have to iterate on all outputs of all transactions like this for each block just to find which transaction has an outpoint with the scripthash.

This is the same when you search a transaction with a certain input, I have to scan all inputs of all transactions of a block to find one that spend an outpoint when implementing outpoint.subscribe in #446.

When you want to get a transaction while knowing the `txid` yeah it is a bit different because Core will find it for us (using the blockhash or not). So you don't even need to ask for the block.

Storing the transaction position in the block avoid all the checks on all other transaction in the same block. So I think it should really be faster !

> However, with your scheme we could just add a single byte which would be 40 bits - 26 for height (same number, yay!), 14 for index. So only 5.55-8.34% increase.
> 
> Just checked that 14 vs 15 bits is just 10 tx difference. To be on the safe side we could store overflow as highest number and continue linear search if tx is not found there.

Great ideas ! Really good :+1:

Kixunil:
Note that one still needs to *iterate* over all txins and txouts in order to even parse the transaction. (there's no transaction size field, so you must decode input count as varint, decode all script **sizes** as varints + do some math to compute next offset then decode output count, decode all output script sizes)

So the additional overhead is just hashing and comparing.

I suggest we write an isolated benchmark that will try to find a transaction as fast as possible using all three methods: seek to position, seek to index, exhaustive search. Also maybe consider only optimizing script finding which requires hashing - input finding only requires comparison, so it could be fast enough. Then we can learn how much it helps.

Pantamis:
`p2p` is merged.

The improvment of this PR was great for old index, with the new one it is less valuable. It will be even less useful when Bitcoin Core 23.0 will be released since the perf loss from specifying blockhash will be resolve at Core level.

The idea is still something interesting for new indexing. In particular, it could be great to add the option to not index `txid` at all with ElectRS if `txindex` is used in Bitcoin Core. This way, we don't lose time in searching the blockhash when we have to get a transaction (without the whole block) from ElectRS.

Personnaly, to improve the perf slightly I simply replaced blockhash by `None` in `get_transaction` function before compiling ElectRS.

I think the goal is now very different from what this PR is doing, so I close it. Thanks for your time on its review @Kixunil !

Kixunil:
> when Bitcoin Core 23.0 will be released since the perf loss from specifying blockhash will be resolve at Core level.

Curious about this, do you have more information?

> Thanks for your time

Thank you for your contribution as well! PR is probably the best thing one can do for an Open Source project. I hope you're enjoying Rust and will see PRs from you again. :)

romanz:
> Curious about this, do you have more information?

IIUC, this is handled by https://github.com/bitcoin/bitcoin/pull/22383.

Pantamis:
> IIUC, this is handled by bitcoin/bitcoin#22383.

Yep, exactly, a bit sad that a such small change (but important !) didn't make it on time for 22.0 release x')

> Thank you for your contribution as well! PR is probably the best thing one can do for an Open Source project. I hope you're enjoying Rust and will see PRs from you again. :)

I have a very busy period to go through first, then I will come back to contribute to bitcoin-rust projects :P with RGB and rust-lightning I think there is a bright futur for Rust usage in Bitcoin ! I want to do more !

Issue #444, opened 2021-07-22
Feature: add blockchain.outpoint Electrum 1.5 methods in p2p branch


Pantamis:
The p2p branch uses a db compatible with version 1.5 of Electrum protocol, one of the new features of 1.5 is the outpoint subscription methods to get the spending transaction of an tx output.

I forked btc-rpc-explorer to add a link to spending transaction of an output to the explorer display using this method, like what blockstream.info explorer does. I would like to test it.

It seems that blockchain.outpoint.subscribe was already implemented in outpoint_RPC branch back in 2018 but didn't made it to p2p. Could you "merge" it with p2p ? (I guess that some glue code may be needed)

Do we need the unsubscribe method in current electrs implementation to properly avoid flood of subscriptions/notifications ?

Thanks !

romanz:
Sounds good, but I prefer adding it when https://github.com/spesmilo/electrumx/pull/90 is officially merged.

Pantamis:
Ok, I understand.

I never coded in rust but is it possible to simply cherry pick the content of the last commit in outpoint-rpc branch and apply it to current p2p one to make it work ?

The files are not the same but taking the diff of last commit of outpoint branch of rpc.rs and apply it on electrum.rs could make it works, didn't it ?

Just a quick way for me to try ^^' (if I have something working I will submit a PR you can merge once Electrum 1.5 is merged too)

romanz:
Please note that the [`outpoint-rpc` branch](https://github.com/romanz/electrs/commits/outpoint-rpc) doesn't implement the notification mechanism specified here:
https://github.com/spesmilo/electrumx/pull/90/files#diff-ffa7e37c618564f8c58d9072b5ef234e6e9b2cdf1d71d16be7dc356146cb475fR648-R655
It was a "WIP" implementation that I put on hold, waiting for the official Electrum 1.5 protocol release.

Pantamis:
Ok, so if I understand well, it will just return the "current" status of the outpoint, that's why you don't need to unsubscribe to notifications.

Pantamis:
Could you take a look at what I tried in #446 ?

I couldn't test it properly for now because I can't increase verbosity with config file and I don't know how to make a request locally to test the method.

Thank for your help !

romanz:
Many thanks for the contribution!
I have pushed a bit different implementation at #454, including the subscription mechanism.

PR #454, opened 2021-08-13
Add blockchain.outpoint.subscribe RPC 

romanz:
Following #444 (see the specification [here](https://github.com/spesmilo/electrumx/pull/90/files#diff-ffa7e37c618564f8c58d9072b5ef234e6e9b2cdf1d71d16be7dc356146cb475fR584)).

Blocked until Electrum client is released with support of 1.5 protocol.

### DIFF
--- a/src/electrum.rs
+++ b/src/electrum.rs
@@ -2,7 +2,7 @@ use anyhow::{bail, Context, Result};
 use bitcoin::{
     consensus::{deserialize, encode::serialize_hex},
     hashes::hex::FromHex,
-    BlockHash, Txid,
+    BlockHash, OutPoint, Txid,
 };
 use crossbeam_channel::Receiver;
 use rayon::prelude::*;
@@ -21,7 +21,7 @@ use crate::{
     merkle::Proof,
     metrics::{self, Histogram, Metrics},
     signals::Signal,
-    status::ScriptHashStatus,
+    status::{OutPointStatus, ScriptHashStatus},
     tracker::Tracker,
     types::ScriptHash,
 };
@@ -36,6 +36,7 @@ const UNSUBSCRIBED_QUERY_MESSAGE: &str = "your wallet uses less efficient method
 pub struct Client {
     tip: Option<BlockHash>,
     scripthashes: HashMap<ScriptHash, ScriptHashStatus>,
+    outpoints: HashMap<OutPoint, OutPointStatus>,
 }
 
 #[derive(Deserialize)]
@@ -185,7 +186,25 @@ impl Rpc {
                 }
             })
             .collect::<Result<Vec<Value>>>()
-            .context("failed to update status")?;
+            .context("failed to update scripthash status")?;
+
+        notifications.extend(
+            client
+                .outpoints
+                .par_iter_mut()
+                .filter_map(|(outpoint, status)| -> Option<Result<Value>> {
+                    match self.tracker.update_outpoint_status(status, &self.daemon) {
+                        Ok(true) => Some(Ok(notification(
+                            "blockchain.outpoint.subscribe",
+                            &[json!([outpoint.txid, outpoint.vout]), json!(status)],
+                        ))),
+                        Ok(false) => None, // outpoint status is the same
+                        Err(e) => Some(Err(e)),
+                    }
+                })
+                .collect::<Result<Vec<Value>>>()
+                .context("failed to update scripthash status")?,
+        );
 
         if let Some(old_tip) = client.tip {
             let new_tip = self.tracker.chain().tip();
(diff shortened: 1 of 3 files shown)


src/electrum.rs
```diff
@@ -266,14 +285,32 @@ impl Rpc {
...
 
-    fn new_status(&self, scripthash: ScriptHash) -> Result<Status> {
-        let mut status = Status::new(scripthash);
+    fn outpoint_subscribe(&self, client: &mut Client, (txid, vout): (Txid, u32)) -> Result<Value> {
+        let outpoint = OutPoint::new(txid, vout);
+        let mut status = OutPointStatus::new(outpoint);
+        self.tracker
+            .update_outpoint_status(&mut status, &self.daemon)?;
+        let result = json!(status);
+        client.outpoints.insert(outpoint, status);
```
Pantamis, line 298: It could be interesting to not subscribe for notifications when the outpoint is spend and the spending transaction has many confirmations (at least 100). This was proposed [here](https://github.com/spesmilo/electrumx/pull/90#discussion_r680886898).

Pantamis:
Code looks very good to me (much better than mine in #446 obviously) and I successfully use it with janoside/btc-rpc-explorer#356

This "test" doesn't use the notification part however !

src/electrum.rs
romanz: Good point - added the following tests to skip index lookups for confirmed funding/spending transactions:
 - https://github.com/romanz/electrs/pull/454/commits/5d2c7cfd26be133fdc0e9701626f3e0970904798#diff-0bfaab7c77d86eb41c2fe32c5e5df509fa02ab04c97b0ce3216911afaae0d67cR527
 - https://github.com/romanz/electrs/pull/454/commits/5d2c7cfd26be133fdc0e9701626f3e0970904798#diff-0bfaab7c77d86eb41c2fe32c5e5df509fa02ab04c97b0ce3216911afaae0d67cR562

romanz:
Thanks for testing1

src/electrum.rs
Pantamis: Ok so if I understand the code well, it will still subscribe for notifications even if the outpoint is spent with many confirmations. However OutpointStatus syncing will only look at the index once, then it only returns its previous value, except if a reorg is detected. If adding notification has no cost if they don't require index lookup, that's nice :)

Pantamis:
Would it be possible to rebase this PR with current p2p branch ?

I would like to try #468 with my PR in the bitcoin explorer janoside/btc-rpc-explorer#356
Thanks :)

romanz:
Thanks for the heads-up!
Rebased aec27af.

Pantamis:
I suppose you will reopen this PR once Electrum 1.5 spec are merged ?

Great work on the `p2p` branch !

Kixunil:
I think this was the same GitHub confusion as in #455

romanz:
> I think this was the same GitHub confusion as in #455

Yes, reopening now :)

Pantamis:
Hi !

Just for information:

This PR still work as expected when cherry picked from master. This is mandatory since #783 

The only thing that must be changed to make it compiling is BlockHash::default() to BlockHash::all_zeros() in OutpointStatus struct.

I re tACK this with janoside/btc-rpc-explorer/pull/356

src/status.rs
```diff
@@ -539,6 +553,131 @@ fn filter_block_txs<T: Send>(
...
+        }
+        let mut confirmed = None;
+        daemon.for_blocks(
+            index.filter_by_txid(self.outpoint.txid),
+            |blockhash, block| {
+                if confirmed.is_none() {
+                    for tx in block.txdata {
+                        let txid = tx.txid();
+                        let output_len = u32::try_from(tx.output.len()).unwrap();
+                        if self.outpoint.txid == txid && self.outpoint.vout < output_len {
```
antonilol, line 630: > `self.outpoint.vout < output_len`

maybe i am missing something, but for what reason is this check here?

Kixunil: Clearly this avoids out-of-bounds access.

antonilol: hmm i tried to find where `self.outpoint.vout` is used to index `tx.output`, but could not find it, `self.outpoint.vout` is not used after this here, which made sense after i thought of this: this code checks for output _funding_, it doesnt matter which output index is subscribed to, as long as it is on the transaction (`self.outpoint.vout < output_len`)

romanz: It's indeed a sanity check - in case the user tries to subscribe to a non-existing outpoint (by specifying a `vout` larger than the actual # of `txid` outputs).

Kixunil: What I'm trying to say is that it now *silently* ignores the transaction instead of printing a warning and a warning would be helpful because doing this is a bug and without a message it could be hard to analyze.

9764319764319764 (approved):

Kixunil:
Funny, this would be useful for me now. I can use script hash and just filter the events of course but not having to do it would be nice.

Issue #423, opened 2021-07-06
MSRV plans regarding bullseye?


Kixunil:
I wonder what the plans regarding bullseye are. Will MSRV be bumped? AFAIK we could do some nice dependency cleanup if it was.

romanz:
> Will MSRV be bumped? AFAIK we could do some nice dependency cleanup if it was.

Yes, I think it will be great to bump to 1.48.0 (https://packages.debian.org/bullseye/rustc) when bullseye is released.

romanz:
Hopefully, it will happen soon™ :)
https://lists.debian.org/debian-release/2021/06/msg00200.html

romanz:
The planned release date is Aug 14:
https://lists.debian.org/debian-devel-announce/2021/07/msg00003.html

Kixunil:
FTR I just successfully compiled `electrs` 0.8.8 (IDK why I still have that one in the project) on bullseye with dynamic linking without issues. Will test later.

Kixunil:
Maybe we should wait a bit longer with MSRV bumping - at least until newer version is in Raspbian, which is quite popular. Or at least do it some time after 0.9.0 release, which will bring important performance improvements and it'd be nice to enable them for RPi users.

My thinking is to release 0.9.0, wait a while to see if there are any issues that should be fixed too, then bump MSRV or bump MSRV when newer version is available in both Debian and Raspbian.

See also https://github.com/Stadicus/RaspiBolt/issues/747#issuecomment-894627169

romanz:
https://bits.debian.org/2021/08/bullseye-released.html :)

romanz:
Latest release bumped MSRV to 1.48.0 (#455).

PR #455, opened 2021-08-15
Update MSRV to 1.48.0 and test on Debian 11

romanz:
Following https://github.com/romanz/electrs/issues/423


Kixunil (approved):

Kixunil:
@romanz do you agree with waiting with this until at least one of these happens?

* 0.9.1 release of `electrs` (assuming non-emergency)
* Raspbian has rustc 1.48

romanz:
Definitely, thanks for the heads up!

Kixunil:
I'm not sure why other commits got mixed here, I believe MSRV bump was planned in 0.10, not 0.9? Also a nit, I think renaming the crate in `Cargo.toml` would have bean cleaner. (And easy to switch back if upstream version is sufficient.)

romanz:
I deleted `p2p` branch, and rebased `p2p-bullseye` over `master` - so I guess GitHub got confused...

romanz:
https://github.com/romanz/electrs/pull/455/checks?check_run_id=3818923414#step:7:90
```
 error: if let .. else expression looks like `matches!` macro
Error:    --> src/p2p.rs:260:5
    |
260 | /     if let Inventory::Block(_) = inv {
261 | |         true
262 | |     } else {
263 | |         false
264 | |     }
    | |_____^ help: try this: `matches!(inv, Inventory::Block(_))`
```

PR #456, opened 2021-08-16
Use the same log verbosity levels as master

Pantamis:
Current verbosity level of master are the ones here: https://docs.rs/stderrlog/0.5.1/src/stderrlog/lib.rs.html#377

p2p branch uses different level values for verbosity d153bfa2eb4435fbe2f70d38ece76152c0bcff66. This surprised me when testing outpoint notification in p2p branch (the trace logs fill my disk storage completely after two days of usage).

This PR restores the same old levels in p2p branch as master.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -268,11 +268,11 @@ impl Config {
         });
 
         let level = match config.verbose {
-            0 => LevelFilter::Error,
-            1 => LevelFilter::Warn,
-            2 => LevelFilter::Info,
-            3 => LevelFilter::Debug,
-            _ => LevelFilter::Trace,
+            0 => log::LevelFilter::Error,
+            1 => log::LevelFilter::Warn,
+            2 => log::LevelFilter::Info,
+            3 => log::LevelFilter::Debug,
+            _ => log::LevelFilter::Trace,
         };
 
         let config = Config {
--- a/src/config.rs
+++ b/src/config.rs
@@ -268,9 +268,11 @@ impl Config {
         });
 
         let level = match config.verbose {
-            0 => log::LevelFilter::Info,
-            1 => log::LevelFilter::Debug,
-            _ => log::LevelFilter::Trace,
+            0 => LevelFilter::Error,
+            1 => LevelFilter::Warn,
+            2 => LevelFilter::Info,
+            3 => LevelFilter::Debug,
+            _ => LevelFilter::Trace,
         };
 
         let config = Config {


Kixunil:
I think it should be done the other way around - change levels in master.

However, I think it'd be even better to extend `configure_me` to accept inverse switch and have default level 2 == info, then `-q` would switch it to warn and `-qq` (or `-q -q`) to error.

Pantamis:
I am not sure about this approach because many raspberry pi nodes use the current level in master. My node has crashed because trace verbosity is very big in p2p branch (much bigger than master, idk why) so we need to really warn users correctly that trace verbosity will occupy lot of space when releasing p2p branch (or reduce trace verbosity before release).

Surprisingly verbose = 4 in the raspibolt guide, which is trace verbosity. So config files will have to be changed anyway if trace verbosity is kept the same in p2p release.

Kixunil:
I think info should be good enough? It's a good point that if we update the default silently it'll cause problems on RPis (but so do the misguided recommendations to set verbosity to trace).

Frankly, I'm not sure `-v` meaning `+ 1` is that great. We could change it to `-v 2` syntax but this is hard-breaking change and difficult to phase-in as softly as we did with `cookie` flag. Maybe we could deprecate it and introduce `--log-level` directly instead. Also make sure to recommend `2` (info; maybe we can also use words to make it more readable). Later when `-v` is removed we could make info default.

Pantamis:
I think it is not that useful to change verbose level in master if next release is the p2p branch.

However this is indeed an opportunity to change how verbosity level works.

> I think info should be good enough?

I think so too !

> but so do the misguided recommendations to set verbosity to trace

What surprise me the most is that it didn't seem to be an issue... until I run p2p-outpoint-rpc branch x'). But yep, we should change the verbosity level on the guides if the electrs part is updated for 0.9.0 release.

I guess everyone has his opinion on what is the best way to allow the user to set verbosity level correctly. This PR only restores the "legacy" (but current) way until another is chosen, if it is the case, we should be really careful on how we inform the users (potentially with depreciated warning as you proposed) and overwrite this PR.

Kixunil:
>  it didn't seem to be an issue

It's a bit like with undefined behavior - it's doesn't seem an issue until it is. Trace log level is always intended to be used for a large number of log messages so advising it for size-constrained production environment is wrong.

Agree with the rest.

romanz:
Let's switch to previous log levels, since now we can use environment variables to customize the logging: https://docs.rs/env_logger/

PR #458, opened 2021-08-18
Limit index lookups to prevent DoS for "popular" addresses

### LINKED ISSUE #447: Config: no txid_limit in p2p branch
ElectRS can hang a long time without responding if it is asked to provide the history of a scripthash with a lot of transactions. txid_limit is used in master branch to stop the query prevently in that case. It doesn't seems to be implemented in p2p branch.

It would be useful to implement Electrum 1.5 compatibility too since get_history will add feature to support pagination of long history (txid_limit could be used to set the length of returned part of history)

romanz:
Fixes #447

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -85,12 +85,18 @@ impl IndexResult {
 /// Confirmed transactions' address index
 pub struct Index {
     store: DBStore,
+    lookup_limit: Option<usize>,
     chain: Chain,
     stats: Stats,
 }
 
 impl Index {
-    pub(crate) fn load(store: DBStore, mut chain: Chain, metrics: &Metrics) -> Result<Self> {
+    pub(crate) fn load(
+        store: DBStore,
+        mut chain: Chain,
+        metrics: &Metrics,
+        lookup_limit: Option<usize>,
+    ) -> Result<Self> {
         if let Some(row) = store.get_tip() {
             let tip = deserialize(&row).expect("invalid tip");
             let headers = store
@@ -103,6 +109,7 @@ impl Index {
 
         Ok(Index {
             store,
+            lookup_limit,
             chain,
             stats: Stats::new(metrics),
         })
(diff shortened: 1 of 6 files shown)


src/index.rs
```diff
@@ -112,6 +119,14 @@ impl Index {
         &self.chain
     }
 
+    pub(crate) fn limit_result<T>(&self, entries: impl Iterator<Item = T>) -> Result<Vec<T>> {
+        let entries: Vec<T> = entries.collect();
```
Kixunil, line 123: I guess it'd be better to use `entries.by_ref().take(self.lookup_limit).collect()` and then check `entries.next().is_some()`? In that case the function should also take `impl Iterator<Item = T> + FusedIterator`.

Kixunil (approved):

Pantamis (approved):
utACK 

I find the @Kixunil suggestion great too ! Should be fairly easy change !

Kixunil:
Had another idea that the value may be nicer to model using `Option<NonZeroUsize>`.

Depends on definition of "nicer". I personally think `Option` expresses non-existing value more clearly and `NonZeroUsize` expresses clearly value required to not be zero which together becomes exactly same as `usize` after optimizations. It also forces refactors to not forget to treat zero value as special.

src/index.rs
romanz: Sounds good, updated the code in https://github.com/romanz/electrs/pull/458/commits/31765e3e49ea93c97721675e2c06e2a9efa76812.
I have called `.fuse()` [here](https://github.com/romanz/electrs/pull/458/commits/31765e3e49ea93c97721675e2c06e2a9efa76812#diff-701633ed548410624cfcae90d38a986adbd60deb5aeedd92355fe2e7702a9c18R123) following the suggestion from https://doc.rust-lang.org/std/iter/trait.FusedIterator.html.

romanz:
Manually tested with [1BitcoinEaterAddressDontSendf59kuE](https://blockstream.info/address/1BitcoinEaterAddressDontSendf59kuE) - which currently has 392 funding transactions from 330 blocks :)

PR #459, opened 2021-08-20
Updated backtrace and relevant dependencies

Kixunil:
This updates `cc`, `memchr` and `backtrace` to hopefully fix #453.
It looks like a memory leak was fixed in a new version but I'm not sure
if relevant now.

See https://github.com/rust-lang/rust/issues/59125#issuecomment-683334971

I didn't test yet, first going to let CI do it.

Issue #453, opened 2021-08-09
Bug: Killed during "starting full compaction"


kentschubert:
electrs Killed during "starting full compaction"

Followed Raspibolt pi set for bitcoin core, lnd, and electrs
$ electrs --conf /mnt/ext/electrs/electrs.conf
Seems to complete indexing...
2021-08-09T02:50:18.499+01:00 - DEBUG - no more blocks to index
2021-08-09T02:50:21.087+01:00 - DEBUG - last indexed block: best=00000000000000000001c35ba10dabccdb8d36789429e0d69d214f068c7ec93a height=694876 @ 2021-08-09T00:24:17Z
2021-08-09T02:50:21.211+01:00 - INFO - starting full compaction
Killed

**Configuration**
AUTH="{myrpcuser}:{myrpcpasswd}"
# Bitcoin Core settings
network = "bitcoin"
daemon_dir= "/mnt/ext/bitcoin"
daemon_rpc_addr = "127.0.0.1:8332"

# Electrs settings
electrum_rpc_addr = "127.0.0.1:50001"
db_dir = "/mnt/ext/electrs/db"
txid_limit = 1000

# Logging
verbose = 4
timestamp = true
rust_backtrace = true

**System running electrs**
raspi 4 8GB + 1.9TB SSD, 4GB swapfile
deployed as per https://stadicus.github.io/RaspiBolt/raspibolt_50_electrs.html#install-dependencies

**Additional context**
Deep into "starting full compaction" system appears to become over-taxed (laggy).
Excerpt from last entries in /mnt/ext/electrs/db/mainnet/LOG :

**Compaction Stats [default]**
Level    Files   Size     Score Read(GB)  Rn(GB) Rnp1(GB) Write(GB) Wnew(GB) Moved(GB) W-Amp Rd(MB/s) Wr(MB/s) Comp(sec) Comp(cnt) Avg(sec) KeyIn KeyDrop
----------------------------------------------------------------------------------------------------------------------------------------------------------
  L0    564/564 90.04 GB   0.0      0.0     0.0      0.0       0.0      0.0       0.0   1.0      0.0      4.3         2         1    2.482       0      0
 Sum    564/564 90.04 GB   0.0      0.0     0.0      0.0       0.0      0.0       0.0   1.0      0.0      4.3         2         1    2.482       0      0
[...]

PR #460, opened 2021-08-20
Backtrace updated to 0.3.55

Kixunil:
This may fix an issue with memory leak but not yet tested.
This change should NOT require MSRV bump.

See #459 and #453

This is an alternative to #459. If this helps then this supersedes #459 else this should be closed.

Yeah, I said I don't have time to try other ways but I figured I could at least try this one. :)


romanz:
Closing since `backtrace` is no longer part of the dependencies.

PR #463, opened 2021-08-23
Move usage in README up and recommend our guide

Kixunil:
* Having usage up should make it more visible.
* Recommend people to use our guide as opposed to external guides which is kept up-to-date with the code.


romanz:
Good idea, thanks!

PR #468, opened 2021-08-31
Run p2p block parsing and processing concurrently

romanz:


src/p2p.rs
```diff
@@ -93,19 +93,31 @@ impl Connection {
...
-        for hash in blockhashes {
-            match self
-                .recv()
-                .with_context(|| format!("failed to get block {}", hash))?
-            {
-                NetworkMessage::Block(block) => {
-                    assert_eq!(block.block_hash(), hash, "got unexpected block");
+
+        rayon::scope(|s| {
+            let (tx, rx) = crossbeam_channel::bounded(1);
```
Kixunil, line 98: Perhaps make it a bit larger (10 or something) to even-out any glitches in processing speed?

Kixunil (approved):
LGTM. I wonder if fetching and processing multiple blocks concurrently would help.

Pantamis:
I am too new to Rust to fully ACK this code but from the little I know in concurrency in Rust it looks good !

src/p2p.rs
```diff
@@ -93,19 +93,31 @@ impl Connection {
...
+                {
+                    NetworkMessage::Block(block) => {
+                        ensure!(block.block_hash() == hash, "got unexpected block");
+                        tx.send((hash, block))
+                            .context("disconnected from block processor")?;
+                    }
+                    msg => bail!("unexpected {:?}", msg),
+                };
+            }
+            drop(tx);
```
Pantamis, line 118: I am quite surprised that you have to use drop method here, I heard you rarely have to call it thanks to ownership rules x')

Kixunil: It's usually used as a comment for the reader if the author considers drop special in some way. In this case that would be "The channel gets closed here which causes stopping the other thread (iterator will return `None`)". It'd be equally valid without it, just less obvious.

Issue #476, opened 2021-09-13
Bug: P2P ignores auth


Kixunil:
**Describe the bug**
P2P version tries to access cookie file despite `auth` being set.

**Electrs version**
P2P branch

**To Reproduce**
1. Configure `electrs` with `auth = "public:public"`
2. Run electrs with `--conf-dir` setting dir where the configuration file is
3. See electrs trying to access cookie inside bitcoin directory

**Expected behavior**
`auth` is used, cookie file is not accessed.

**Configuration**

<details>
<summary>electrs.toml</summary>

```
# Automtically generated - DO NOT MODIFY!

db_dir="/var/lib/electrs-regtest"

jsonrpc_import = true

verbose = 4

auth="public:public"

daemon_dir="/var/lib/bitcoin-regtest"

daemon_rpc_addr="127.0.0.1:18443"

network="regtest"

electrum_rpc_addr="127.0.0.1:60401"

monitoring_addr="127.0.0.1:24224"
```

</details>

Environment variables:
Arguments: `--conf-dir /etc/...`

**System running electrs**
 - Deployment method: manual
 - OS name and version: Debian 10

**Electrum client**
None

**Additional context**
Trying to test auto reindex feature, will send PR anyway soon.

Kixunil:
Looks like I got old p2p branch, not sure why it didn't update.

PR #477, opened 2021-09-13
Auto reindex when old database is detected

Kixunil:
This implements automatic reindexing of old database, turned on by
default. When an old database is detected and `auto_reindex` is turned
on the database will be destroyed so that `electrs` can re-sync. The
user can still turn this off to check db format.

The help message says "or inconsistent" - this is not implemented but we
could in the future and it seems reasonable to explicitly state that
this option will control that thing as well.

The code was also restructured a bit, biggest changes were due to `Drop`
impl.

Not tested yet due to #476 Will try to workaround soon.


Kixunil:
This is broken because database format detection logic is broken. `C` was introduced in `p2p` and there are two scenarios when it may be missing:

* Fresh new DB
* Old format

I think the fix is to assume current version if *no record* exists (empty db) and old if there's anything. Will try to implement it soon.

src/db.rs
```diff
@@ -89,27 +100,46 @@ impl DBStore {
             live_files.iter().map(|f| f.size).sum::<usize>() as f64 / 1e9,
             live_files.iter().map(|f| f.num_entries).sum::<u64>() as f64 / 1e9
         );
+        let mut config = Self::get_config(&db);
+        debug!("DB {:?}", config);
+        if config.format != CURRENT_FORMAT {
+            if auto_reindex {
+                info!("Database needs to be reindexed due to being in old format, going to delete the old one.");
```
romanz, line 107: nit:
```suggestion
                info!("Database needs to be reindexed due to being in unsupported format ({} != {}), going to delete {}", config.format, CURRENT_FORMAT, path.display());
```

romanz (commented):
Many thanks, a few comments below

src/db.rs
```diff
@@ -42,6 +41,8 @@ const TXID_CF: &str = "txid";
 const FUNDING_CF: &str = "funding";
 const SPENDING_CF: &str = "spending";
 
+const CFS: &[&str] = &[CONFIG_CF, HEADERS_CF, TXID_CF, FUNDING_CF, SPENDING_CF];
```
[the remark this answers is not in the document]
Kixunil: Finally I understand what it means :laughing:

src/db.rs
```diff
@@ -68,19 +78,20 @@ fn default_opts() -> rocksdb::Options {
...
 
+    /// Opens a new RocksDB at the specified location.
+    pub fn open(path: &Path, auto_reindex: bool) -> Result<Self> {
         let mut db_opts = default_opts();
         db_opts.create_if_missing(true);
         db_opts.create_missing_column_families(true);
-        let db = rocksdb::DB::open_cf_descriptors(&db_opts, path, cf_descriptors)
-            .with_context(|| format!("failed to open DB: {:?}", path))?;
+        let mut db =
+            rocksdb::DB::open_cf_descriptors(&db_opts, path, Self::create_cf_descriptors())
```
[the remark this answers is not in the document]
Kixunil: Oh, that's a good way to distinguish the versions! Will need to distinguish it from OS errors, hopefully it can be done.

src/db.rs
Kixunil: Nice!

src/db.rs
```diff
@@ -89,27 +100,46 @@ impl DBStore {
             live_files.iter().map(|f| f.size).sum::<usize>() as f64 / 1e9,
             live_files.iter().map(|f| f.num_entries).sum::<u64>() as f64 / 1e9
         );
+        let mut config = Self::get_config(&db);
+        debug!("DB {:?}", config);
+        if config.format != CURRENT_FORMAT {
```
Kixunil, line 105: Actually realized this will reindex on downgrades too. I don't think we want it, do we?

I'm thinking of changing `auto_reindex` to some sort of bit flags (how do we deserialize them?):

* Upgrades - `config.format` < `CURRENT_FORMAT`
* Downgrades - `config.format` > `CURRENT_FORMAT`
* Inconsistencies - everything else
* Always - like `-reindex` in bitcoind - forces reindex; perhaps a separate argument?

Default is Upgrades + Inconsistencies

Kixunil: Perhaps could be specified as comma-separated list:

* CLI: `--reindex=upgrades,downgrades,inconsistencies`
* config: `reindex = "upgrades,downgrades,inconsistencies"`

Not sure what `bitflags` crate does to deserialize.

romanz:
I think we can add Snappy to make the check easier -> https://github.com/romanz/electrs/commit/b14fe1a1b7c4605c245f1cdfd3130ab6d5aecf44.

Kixunil:
Did you intend to make a [PR](https://github.com/romanz/electrs/compare/p2p...p2p-snappy)? Or should I base my changes on `p2p-snappy`?

src/db.rs
romanz: Let's keep it simple for now - and re-index every time the DB format changes :)

romanz:
Merged snappy support into `p2p` - please rebase :)

src/db.rs
Kixunil: I wouldn't mind writing flags but OK. :)

Kixunil:
@romanz for some unknown reason I'm unable to produce legacy database on regtest. Could you test it please? I'm fairly confident in the logic and the change is not too complex so shouldn't be too much hassle. :)

Kixunil:
(sorry for double brain fart)

romanz:
Looks good, thanks!
I have adapted the implementation a bit here:
https://github.com/romanz/electrs/compare/p2p...p2p-auto-reindex
WDYT?

Kixunil:
Veeery nice!!!

Do I understand correctly that `.iterator(rocksdb::IteratorMode::Start)` will **not** return any entries for database using CF?

romanz:
> Veeery nice!!!

Thanks :)

> Do I understand correctly that `.iterator(rocksdb::IteratorMode::Start)` will **not** return any entries for database using CF?

Yes, `db.iterator()` only scans the `default` column famility - and since the new DB index uses only non-`default` CFs, the `default` one should be empty.

Kixunil:
Great! Your code looks correct and better than mine, so I'm closing this in favor of your version.

romanz:
Cool, merged via 577f54195b8c21f63fc4abec3c9364ddd1a73692.

Issue #371, opened 2021-02-22
Tracking issue: deprecation and removal of `cookie` option


Kixunil:
This tracks deprecation and removal of `cookie` option which was confusing.

- [x] Implement deprecation of `cookie` - warning in the code and explanation in docs (#360)
- [x] Release a version with deprecation
- [x] Turn warning into hard error
- [x] Remove all code handling `cookie` option

Kixunil:
@romanz Do we want to do last two steps as one in 0.9 or should we revert warning removal and turn it into error?

romanz:
Let's drop `cookie` in `0.9.0-rc1` - so the user will get the following error:
```
Error: An unknown argument '--cookie' was specified
```
WDYT?

Kixunil:
Yeah, originally I intended for it to be more clearly state that this argument was known in the past but was removed and point to docs. But if you think it's not worth the effort I'm fine with it being removed entirely (which it currently is).

romanz:
You're right, let's add an explicit failure in case `cookie` is specified at 0.8.12 -> #478

romanz:
0.8.12 will be released soon: #480

Kixunil:
So effectively 0.8.12 only disables cookie and has no other change. Does it make sense to release it?

romanz:
It would be the last (and the smallest ever) 0.8.x release :)

romanz:
IIUC, this issue can be closed since there is no `cookie`-relevant code left in `master` :)

Engelberg:
I have never used the cookie option, nor do I have a cookie file, but when I upgraded to version 0.9.4, I get the following error:

```
Error: electrs failed

Caused by:
    0: failed to open bitcoind cookie file: /home/mark/.bitcoin/.cookie
    1: No such file or directory (os error 2)
```

romanz:
Could you please check that `bitcoind` is running?

Engelberg:
bitcoind is running, but it looks like maybe it's trying to reach it at port 8333, possibly not respecting my flag --daemon-rpc-addr=127.0.0.1:28256. Has that configuration option changed recently?

Engelberg:
Ah, I see port 8333 is the p2p port not the rpc port. I don't know why it would be having trouble reaching that, but that does seem to be a factor here. I will keep poking around.

Engelberg:
Needed to set the daemon_p2p_addr. Not sure what changed between versions, but this fixed my problem.

[...]

PR #478, opened 2021-09-13
Fail if cookie is specified

romanz:
Following https://github.com/romanz/electrs/issues/371#issuecomment-918217030


Kixunil:
LGTM

Issue #475, opened 2021-09-13
Feature: blockchain.scripthash.listunspent


gr0kchain:
Electrum v1.3 deprecates `blockchain.address.listunspent` in favour of `blockchain.scripthash.listunspent`.

Noted that this is currently available in the master branch, but not the p2p branch which includes batching support. 

https://electrumx.readthedocs.io/en/latest/protocol-methods.html#blockchain.scripthash.listunspent

romanz:
Please take a look at #485 :)

romanz:
Should be now resolved in `p2p` branch.

gr0kchain:
Tested and this seems to work as expected! Thanks for taking the time to assist in getting this implemented

PR #483, opened 2021-09-15
Simplify 'blockchain.scripthash.get_balance' implementation 

romanz:
No need to use transaction cache for getting output values.


romanz:
It should also help with implementing #475.

PR #484, opened 2021-09-15
Split CI into build and integration jobs

romanz:
Use debug build for build/clippy/fmt/unittests (using https://github.com/actions-rs).
Use release build for integration tests (using Docker).


Kixunil (approved):
Looks good. Perhaps also add `-D warnings`?

Issue #494, opened 2021-09-20
Bug: initial sync do not support graceful exit using SIGINT


romanz:
**Describe the bug**
Since initial sync doesn't use [WAL](https://github.com/facebook/rocksdb/wiki/Write-Ahead-Log) when writing to rocksdb (for performance reasons) and signal handlers are registered after the initial sync is over, sending a SIGINT/SIGTERM results in an empty index DB.
Therefore, interrupting the initial sync will require a full resync (from genesis block).

**Electrs version**
0.9.0-rc1

**Expected behavior**
We should resume from the last block indexed and successfully written to the DB.

Kixunil:
0.10 milestone? This sounds like a pretty sad bug that may be reasonably fixed in 0.9. Is there some difficulty?

romanz:
Good point - I'll try to do it this week :)

romanz:
Should be fixed in 0.9.0-rc2 :)

PR #498, opened 2021-09-21
Close index DB if interrupted during initial sync

romanz:
It should allow resuming initial sync (fixing #494).

Also, refactor `index_batch_size` usage.

### DIFF
--- a/src/signals.rs
+++ b/src/signals.rs
@@ -3,30 +3,72 @@ use crossbeam_channel::{unbounded, Receiver};
 use signal_hook::consts::signal::*;
 use signal_hook::iterator::Signals;
 
+use std::sync::{
+    atomic::{AtomicBool, Ordering},
+    Arc,
+};
+
 use crate::thread::spawn;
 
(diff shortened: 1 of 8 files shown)


src/signals.rs
```diff
@@ -3,30 +3,72 @@ use crossbeam_channel::{unbounded, Receiver};
...
-        Ok(())
-    });
-    rx
+    }
+
+    pub fn get(&self) -> bool {
+        self.flag.load(Ordering::Relaxed)
+    }
+
+    fn set(&self) {
```
Kixunil, line 29: It's not entirely clear from the method names whether it means "exiting" or "running" (although type name could be considered sufficient?) Perhaps name them `is_exiting()` and `set_exiting()`

Kixunil (approved):
Looks good

src/signals.rs
romanz: Renamed `get` into `is_set`, keeping `set` as is - since it's private.

PR #501, opened 2021-09-24
Use blocks' height iterator (instead of a map)

romanz:
We already assume that blocks are processed in ascending order.

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -2,8 +2,6 @@ use anyhow::Result;
 use bitcoin::consensus::{deserialize, serialize};
 use bitcoin::{Block, BlockHash, OutPoint, Txid};
 
-use std::collections::HashMap;
-
 use crate::{
     chain::Chain,
     daemon::Daemon,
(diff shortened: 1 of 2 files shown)


src/index.rs
```diff
@@ -182,17 +180,15 @@ impl Index {
...
-
-                daemon.for_blocks(blockhashes, |blockhash, block| {
-                    let height = heights_map.remove(&blockhash).expect("unexpected block");
+                daemon.for_blocks(blockhashes, |_blockhash, block| {
+                    let height = heights.next().expect("unexpected block");
                     let result = index_single_block(block, height);
                     result.extend(&mut batch);
                 })?;
-                assert!(heights_map.is_empty(), "some blocks were not indexed");
+                assert_eq!(heights.next(), None, "some blocks were not indexed");
```
Kixunil, line 191: Since presumably you use `assert_eq!` instead of `is_none()` to get debug output containing the hash of not-indexed block maybe use `heights.collect::<Vec<_>>()` and compare to `Vec::new()`? That way we have no allocation in happy path and learn all extra hashes in debug output if things go wrong.

Or, if you want to get rid of allocations completely, use `chunks_mut` and `drain(..)` instead of `iter()`, then just check if chunk is empty at the end. :)

src/index.rs
```diff
@@ -182,17 +180,15 @@ impl Index {
...
-                let mut heights_map: HashMap<BlockHash, usize> =
-                    chunk.iter().map(|h| (h.hash(), h.height())).collect();
+                let mut heights = chunk.iter().map(|h| h.height());
 
                 let mut batch = WriteBatch::default();
-
-                daemon.for_blocks(blockhashes, |blockhash, block| {
-                    let height = heights_map.remove(&blockhash).expect("unexpected block");
+                daemon.for_blocks(blockhashes, |_blockhash, block| {
+                    let height = heights.next().expect("unexpected block");
```
Kixunil, line 187: Not sure if panicking because external input is unexpected is the best course of action but I find it understandable if you believe it is.

Kixunil (approved):
Logic looks good, just two design notes/ideas.

Pantamis (approved):
LGTM :+1:

src/index.rs
romanz: Sounds good, changed to `collect`.

romanz:
Thanks :)

Issue #502, opened 2021-09-24
Bug: Electrs 0.9.0-rc1 uses overall more storage than less when reindexing


dematerializer:
Sorry for being late with this bug report, but reindexing a previously fully indexed electrs `0.8.x` db with `0.9.0-rc1` results in two `electrs/db` subdirectories: `electrs/db/mainnet` (~76GB) **and** `electrs/db/bitcoin` (~30GB).

```
admin@raspberrypi:~ $ du -h /mnt/ext/electrs/db
76G	/mnt/ext/electrs/db/mainnet
30G	/mnt/ext/electrs/db/bitcoin
106G	/mnt/ext/electrs/db
```

Is this normal behaviour? Because this is contrary to the release tweet stating that [Index DB uses ~60% less storage](https://twitter.com/roman_zeyde/status/1439173165163425792) but in the end it uses **more** storage overall. Maybe it should be mentioned in the [Upgrading section of the usage guide](https://github.com/romanz/electrs/blob/master/doc/usage.md#upgrading) to manually delete the old `electrs/db/mainnet` directory?

dematerializer:
I already deleted the old `electrs/db` directory and indexed from scratch with `0.9.0-rc2`, so I would have a hard time reproducing this on my node again.

romanz:
Many thanks for reporting this issue!
`mainnet` was the old name for the DB directory in 0.8.x releases.
It was indeed changed in 0.9 - and I forgot to update the release notes :)

romanz:
It should be safe to delete the old `db/mainnet` directory after upgrading to 0.9.x release.

PR #503, opened 2021-09-25
Add a release note about `db/mainnet` -> `db/bitcoin` rename

romanz:
https://github.com/romanz/electrs/issues/502

### DIFF
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -498,6 +498,7 @@ Upgrading checklist:
   Because of reindex electrs will be unable to serve your requests for a few hours.
   (The exact time depends on your hardware.)
   If you wish to check the database without reindexing run electrs with `--no-auto-reindex`.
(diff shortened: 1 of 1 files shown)


Kixunil:
So if I understand this correctly the code in #477 is bogus because it will create a new directory anyway? If yes, then we should delete it.

romanz:
It's useful for testnet/signet and will be useful for next upgrades :)

Kixunil:
I mean we should delete the directory, not the code. :)

romanz:
You're right, but fixing this would require us to test both `db/mainnet` and `db/bitcoin` - complicating the upgrade code :(

Kixunil:
Hmm, I don't think complication will be *that* bad. Will try to look into it soon but my time is not great lately so it may take two-three days.

PR #504, opened 2021-09-25
Warn about log level

Kixunil:
Some people still incorrectly use trace log level and get burned by this update.


romanz (approved):
Many thanks!

Issue #495, opened 2021-09-21
Bug: tip_loop thread failed: JSON-RPC error: transport error: Didn't receive response data in time, timed out.


openoms:
Reproduced on both arm64 RaspberryOS  (Debian base) and amd64 Ubuntu 20.04  VM both using RaspiBlitz scripts.
The install script is developed and discussed here: https://github.com/rootzoll/raspiblitz/issues/2548
Running 0.9.0-rc1 and same on master.

indexed and compacted successfully, but since then Electrs keeps restarting with the error:
```
WARN  electrs::thread] tip_loop thread failed: JSON-RPC error: transport error: Didn't receive response data in time, timed out.
```
Bitcoind is running without errors.

More logs:
```
systemd[1]: Stopped Electrs.
systemd[1]: Started Electrs.
electrs[1167816]: Config { network: Bitcoin, db_path: "/mnt/hdd/app-storage/electrs/db/bitcoin", daemon_dir: "/home/electrs/.bitcoin", daemon_auth: UserPass("raspibolt", "<sensitive>"), daemon_rpc_addr: 127.0.0.1:8332, daemon_p2p_addr: 127.0.0.1:8333, electrum_rpc_addr: 127.0.0.1:50001, monitoring_addr: 127.0.0.1:4224, wait_duration: 20s, index_batch_size: 10, index_lookup_limit: Some(200), auto_reindex: true, ignore_mempool: false, sync_once: false, server_banner: "Welcome to electrs 3041e89cd2fb377541b929d852ef6298c2d4e60a - the Electrum Rust Server on your RaspiBlitz", args: [] }
electrs[1167816]: [2021-09-21T08:43:40.686Z INFO  electrs::metrics::metrics_impl] serving Prometheus metrics on 127.0.0.1:4224
electrs[1167816]: [2021-09-21T08:43:40.984Z INFO  electrs::db] "/mnt/hdd/app-storage/electrs/db/bitcoin": 129 SST files, 31.649062994 GB, 3.94259865 Grows
electrs[1167816]: [2021-09-21T08:43:43.019Z INFO  electrs::chain] loading 701538 headers, tip=0000000000000000000be09a4d70c768038a3b7e153517276e9d04e1fb615c48
electrs[1167816]: [2021-09-21T08:43:44.904Z INFO  electrs::chain] chain updated: tip=0000000000000000000be09a4d70c768038a3b7e153517276e9d04e1fb615c48, height=701538
[...]

PR #509, opened 2021-09-27
Don't fail on HTTP timeouts from bitcoind RPC

romanz:
It should be OK to retry in case of a RPC timeout.

Should help with https://github.com/romanz/electrs/issues/495.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -47,6 +47,21 @@ impl Peer {
     }
 }
 
+fn handle_rpc_error(name: &str, err: bitcoincore_rpc::Error) -> Result<()> {
+    use bitcoincore_rpc::{
+        jsonrpc::{error::Error::Transport as TransportError, simple_http::Error as HttpError},
+        Error::JsonRpc as JsonRpcError,
+    };
+    if let JsonRpcError(TransportError(ref e)) = err {
+        if let Some(HttpError::Timeout) = e.downcast_ref::<HttpError>() {
+            // Following https://github.com/romanz/electrs/issues/495
+            warn!("ignoring bitcoind HTTP timeout error");
+            return Ok(());
+        }
+    }
+    bail!("RPC '{}' failed: {}", name, err); // fail on all other RPC errors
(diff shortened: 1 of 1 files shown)


Kixunil (approved):
`match` would feel more rust-idiomatic but logic looks good.

romanz:
Thanks!

> `match` would feel more rust-idiomatic but logic looks good.

I agree, but clippy does not :)
https://github.com/romanz/electrs/runs/3723483967?check_suite_focus=true#step:7:89
https://rust-lang.github.io/rust-clippy/master/index.html#single_match

Kixunil:
Was thinking of moving `bail!` into the `_ =>` arm, I believe it should not trigger `clippy` (and if it does I would probably report it as a bug).

romanz:
Not sure if I understand... did you mean:
```rust
    match err {
        JsonRpcError(TransportError(ref e)) if let Some(HttpError::Timeout) = e.downcast_ref::<HttpError>() => {
            // Following https://github.com/romanz/electrs/issues/495
            warn!("ignoring bitcoind HTTP timeout error");
            return Ok(());
        }
        _ => bail!("RPC '{}' failed: {}", name, err), // fail on all other RPC errors
    }
```
Unfortunately, the above code fails to compile with:
```
error[E0658]: `let` expressions in this position are experimental
  --> src/server.rs:56:48
   |
56 |         JsonRpcError(TransportError(ref e)) if let Some(HttpError::Timeout) = e.downcast_ref::<HttpError>() => {
   |                                                ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
   |
   = note: for more information, see https://github.com/rust-lang/rust/issues/53667
```

Kixunil:
Ah, I see the problem now, didn't realize `bail!` would have to be repeated in `else` branch and your example is not supported on stable Rust. OK, let's keep it as it was.

Issue #505, opened 2021-09-25
Issue determining index height in v0.9.0-rc2


tehelsper:
**Describe the bug**
I started testing the release candidate of electrs in myNode and got it running. There is some logic within myNode that attempts to track the current block during electrs initial sync to provide users with a status so they can see electrs is running and working properly. It was determining this by polling the Prometheus data every so often. With the latest RC, the Prometheus data seems very different than before and current block is no longer available. Was this intentional?

**Electrs version**
Testing v0.9.0-rc2

**To Reproduce**
1. Start electrs with Prometheus data enabled
2. Check Prometheus data during sync
3. `electrs_index_height` is missing in the data

**Expected behavior**
I expected to see the following line in the Prometheus data.

`electrs_index_height 307979`

But this is all I get.

```
admin@mynode:~$ curl http://localhost:4224
# HELP index_update_size Index update size (in bytes)
# TYPE index_update_size histogram
index_update_size_bucket{step="write_funding_rows",le="0.005"} 0
index_update_size_bucket{step="write_funding_rows",le="0.01"} 0
index_update_size_bucket{step="write_funding_rows",le="0.025"} 0
index_update_size_bucket{step="write_funding_rows",le="0.05"} 0
index_update_size_bucket{step="write_funding_rows",le="0.1"} 0
index_update_size_bucket{step="write_funding_rows",le="0.25"} 0
index_update_size_bucket{step="write_funding_rows",le="0.5"} 0
index_update_size_bucket{step="write_funding_rows",le="1"} 0
index_update_size_bucket{step="write_funding_rows",le="2.5"} 0
index_update_size_bucket{step="write_funding_rows",le="5"} 0
index_update_size_bucket{step="write_funding_rows",le="10"} 0
index_update_size_bucket{step="write_funding_rows",le="+Inf"} 1556
index_update_size_sum{step="write_funding_rows"} 1278818184
index_update_size_count{step="write_funding_rows"} 1556
index_update_size_bucket{step="write_header_rows",le="0.005"} 0
index_update_size_bucket{step="write_header_rows",le="0.01"} 0
index_update_size_bucket{step="write_header_rows",le="0.025"} 0
index_update_size_bucket{step="write_header_rows",le="0.05"} 0
index_update_size_bucket{step="write_header_rows",le="0.1"} 0
index_update_size_bucket{step="write_header_rows",le="0.25"} 0
index_update_size_bucket{step="write_header_rows",le="0.5"} 0
[...]

PR #515, opened 2021-09-30
Initialize chain height metric

romanz:
Otherwise, it will be 0 until the a new block is indexed.

Following https://github.com/romanz/electrs/issues/505#issuecomment-930651597.


romanz:
CC: @tehelsper

Kixunil (approved):
LGTM

PR #518, opened 2021-09-30
Update documentation regarding cargo features

### LINKED ISSUE #517: Bug: Wrong documentation of `--no-default-features`
[usage.md#dynamic-linking](https://github.com/romanz/electrs/blob/34a8549d2e436c52a394e4c35169c9b5efc65fe9/doc/usage.md#dynamic-linking) recommends to enable dynamic linking with `--no-default-features` and states that  `--no-default-features [is] only related to rocksdb linking`. 
But in version 0.9.0 this also disables the `metrics` feature. How does one enable both dynamic linking and `metrics`?

Kixunil:
With new `rocksdb` it's no longer necessary to disable features when dynamically linking.
Thus this change removes instructions to add `--no-default-features`.
However there's now a new feature called `metrics` which enables Prometheus.
This change also documents this and explains how to turn it off.

Closes #517

### DIFF
--- a/doc/usage.md
+++ b/doc/usage.md
@@ -96,6 +96,12 @@ Note: you need to have enough free RAM to build `electrs`.
 The build will fail otherwise.
 Close those 100 old tabs in the browser. ;)
 
+#### Cargo features
+
+By default `electrs` builds with Prometheus support.
+However this causes problems on some platforms.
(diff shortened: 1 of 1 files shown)


romanz (approved):
Thanks!

Issue #508, opened 2021-09-27
Config: "blockchain.scripthash.get_balance called for unsubscribed scripthash"


aitorjs:
I'm calling to `blockchain.scripthash.get_balance` from netcat like this:

```bash
echo '{"jsonrpc": "2.0", "method": "blockchain.scripthash.get_balance", "params": ["93df2ed6d3031610206a9ef6ee5d7c2bc538d6689ffb5ce964144b2b90bdc9e1"], "id": 1}' | netcat 127.0.0.1 60001
```

I get the result but I get this warning in the console:

```bash
[2021-09-27T16:17:04.902Z WARN  electrs::electrum] blockchain.scripthash.get_balance called for unsubscribed scripthash: 93df2ed6d3031610206a9ef6ee5d7c2bc538d6689ffb5ce964144b2b90bdc9e1
```
[this line of code](https://github.com/romanz/electrs/blob/master/src/electrum.rs#L242)

If I call from javascript using netcat, I get the warning commented before and this error:
```bash
[2021-09-27T16:19:04.995Z ERROR electrs::server] 13: disconnecting due to failed to send response: "{\"id\":0,\"jsonrpc\":\"2.0\",\"result\":{\"confirmed\":0,\"unconfirmed\":0}}\n"
```
so client data event is never called and I dont get the answer.

The javascript code I'm using to call `blockchain.scripthash.get_balance` runs ok on [this](https://github.com/romanz/electrs/commit/5854b20811044b1dd84dc32eaef7b0b0a642dbc6) commit.

I try to subscribe to scripthash 93df2ed6d3031610206a9ef6ee5d7c2bc538d6689ffb5ce964144b2b90bdc9e1 like this:

```bash
$ echo '{"jsonrpc": "2.0", "method": "blockchain.scripthash.subscribe", "params": ["93df2ed6d3031610206a9ef6ee5d7c2bc538d6689ffb5ce964144b2b90bdc9e1"], "id": 1}' | netcat 127.0.0.1 60001

> {"id":1,"jsonrpc":"2.0","result":"f7dd2df65f0eb38f413b46c242a9b11da184aada073d797fde3ce32fb33f107e"}
```

I try netcat to get_balance but I still get the warning. 

¿What I'm missing? 
Thanks! :)

Kixunil:
Hi, could you please fill the issue template? We are missing crucial information such as *electrs version* which is required to investigate.

romanz:
Many thanks for reporting this issue!
[...]

PR #523, opened 2021-09-30
Don't shutdown write-side before all responses are sent back

romanz:
Otherwise, the server fails to send RPC result to the client:
https://github.com/romanz/electrs/issues/508#issuecomment-928254892

The bug can be reproduced on testnet using:
```python
$ cat test.py 
import socket
import sys
import time

s = socket.create_connection(('localhost', 60001))
s.sendall(b'{"jsonrpc": "2.0", "method": "blockchain.scripthash.subscribe", "params": ["93df2ed6d3031610206a9ef6ee5d7c2bc538d6689ffb5ce964144b2b90bdc9e1"], "id": 1}\n')
s.shutdown(socket.SHUT_WR)  # the server should stop "receiver loop" now, but not drop the connection

time.sleep(1)
print(s.recv(1000))
```
The server fails to send the response:
```
[2021-09-30T17:51:14.361Z ERROR electrs::server] 0: disconnecting due to failed to send response: "{\"id\":1,\"jsonrpc\":\"2.0\",\"result\":\"f7dd2df65f0eb38f413b46c242a9b11da184aada073d797fde3ce32fb33f107e\"}\n"
```

After this fix, the response should be received correctly by the client.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -43,7 +43,9 @@ impl Peer {
     }
 
     fn disconnect(self) {
-        let _ = self.stream.shutdown(Shutdown::Both);
+        if let Err(e) = self.stream.shutdown(Shutdown::Both) {
+            warn!("{}: failed to shutdown TCP connection {}", self.id, e)
+        }
     }
 }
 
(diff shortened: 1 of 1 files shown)


Kixunil (approved):
Looks correct. Strictly unrelated but shouldn't we at least log the error?

romanz:
Good point, thanks!

Issue #522, opened 2021-09-30
Suggestion: Consider future use of `waitfornewblock`


erikarvstedt:
Since version 0.9.0, electrs uses the `waitfornewblock` RPC fn (via [rust-bitcoincore-rpc](https://github.com/rust-bitcoin/rust-bitcoincore-rpc) method `wait_for_new_block`) which is marked as hidden and reserved for testing. It was once [consided for removal](https://github.com/bitcoin/bitcoin/pull/10519).

This might brake deployments that use RPC whitelists which only include public calls.
The implementation might also have robustness issues as mentioned [here](https://github.com/bitcoin/bitcoin/pull/10519#issuecomment-323094194).

Electrs should or either stop using this fn or lobby to make it public. Nicolas Dorier is also [very interested](https://github.com/bitcoin/bitcoin/pull/10519#issuecomment-371250615) in using it for btcpayserver.

Kixunil:
Thanks for heads up! That seems to make three projects using them. Would like to look into making them public.

romanz:
Maybe we can use p2p inv messages (announcing new blocks) instead of relying on a hidden RPC.
Please take a look at https://github.com/romanz/electrs/compare/p2p-waitfornewblock :)

Kixunil:
If I understand the code correctly two things (receiver loop and block reader) may attempt to read from same connection at the same time leading to lost or corrupt messages. I'd also like to see `Option<Sender<()>>` without `Option`.

Will try to think of a safer design.

Kixunil:
Oh, it looks like it creates another connection? That would fix sync issue but it looks like bitcoind doesn't like too many connections...

romanz:
> Oh, it looks like it creates another connection?

Yes - it was the simplest fix :)

> but it looks like bitcoind doesn't like too many connections...

I agree, we should reuse the same p2p connection for both getting headers/block data and notifications.

romanz:
Should be resolved in `master`.

PR #526, opened 2021-10-03
Use p2p protocol to replace `waitfornewblock` hidden RPC

romanz:
Fixes https://github.com/romanz/electrs/issues/522.

### DIFF
--- a/src/daemon.rs
+++ b/src/daemon.rs
@@ -4,6 +4,7 @@ use bitcoin::{
     consensus::serialize, hashes::hex::ToHex, Amount, Block, BlockHash, Transaction, Txid,
 };
 use bitcoincore_rpc::{json, jsonrpc, Auth, Client, RpcApi};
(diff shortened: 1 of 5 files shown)


romanz:
[... with entirely too many Crossbeam Channels](https://rust-unofficial.github.io/too-many-lists/) :)

romanz:
> So if we're unlucky and two threads are waiting for a response the scheduler could wake up the thread that sent the latter request first and that thread would presumably receive the response intended for the other.

Yes, it's indeed currently assumed that `get_new_headers` and `for_blocks` can be called only from one thread (since they are using `&mut self`).

Kixunil:
Ah, yes, `&mut` protects it. So I believe the code is OK, just a bit hazardous in case of refactors?

romanz:
Indeed... I'll add a comment about `&mut` and `new_block_notification` expected usage.

PR #527, opened 2021-10-04
Update build description for 32-bit OS

romanz:
Following #134 and #391.


xanoni (approved):
Yup, that's what works for me!

PR #559, opened 2021-10-15
contrib/xpub.py support ypub/zpub

flepied:


romanz:
Many thanks, merged via b114826.

Issue #575, opened 2021-10-19
`cargo audit` fails due to RUSTSEC-2020-0159


romanz:
https://github.com/chronotope/chrono/issues/499

romanz:
https://github.com/romanz/electrs/runs/3933536627

romanz:
https://www.reddit.com/r/rust/comments/qamgyh/is_the_chrono_crate_unmaintained

Kixunil:
Hmm, not a huge issue since we don't call `setenv`. We only need chrono for formatting logs, right?

Kixunil:
Maybe worth looking into removing time like they did here: https://github.com/rusqlite/rusqlite/pull/1031
IDK if we need it or not.

Kixunil:
Looks like the fix is in 0.9 https://github.com/tiny-http/tiny-http/commit/a8260ff12a7d491e6251372324eb5dd513fbced8

PR #576, opened 2021-10-19
Update tiny_http and drop `cargo audit` CI for now

romanz:
Following #575.


romanz:
Not sure how to do that with https://github.com/actions-rs/audit-check :(
I tried adding `args`, but it didn't recognize them:
https://github.com/romanz/electrs/runs/3942930966

romanz:
Cool, it seems to work :)
https://github.com/romanz/electrs/runs/3949291782#step:3:16

Issue #573, opened 2021-10-18
Bug: lenghty initialization on each start


schildbach:
**Describe the bug**
After updating from 0.8.x, to 0.9.0 and then 0.9.1, electrs goes through a lengthy initialization, producing these two messages in the log over and over again:

```
electrs_1           | [2021-10-18T15:57:48.619Z DEBUG electrs::p2p] loading 10 blocks
electrs_1           | [2021-10-18T15:57:48.622Z DEBUG electrs::index] writing 18 funding and 4 spending rows from 14 transactions, 10 blocks
```

The process takes roughly an hour for testnet. That's fine, as it is mentioned in the release notes. However it's unexpected this process needs to be repeated over and over again for each restart of electrs.

The old 0.8.x. version also had some initialization time, but it was only about 1-2 minutes.

**Electrs version**
0.9.1 (and 0.9.0)

**To Reproduce**
1. Configure and start electrs

**Expected behavior**
Init times < 2 minutes for testnet. 

**Configuration**

```
      - ELECTRS_NETWORK=testnet
      - ELECTRS_VERBOSE=3
      - ELECTRS_ELECTRUM_RPC_ADDR=0.0.0.0:…
      - ELECTRS_DAEMON_RPC_ADDR=bitcoind:…
      - ELECTRS_DAEMON_P2P_ADDR=bitcoind:…
      - ELECTRS_COOKIE_FILE=…
      - ELECTRS_WAIT_DURATION_SECS=10
      - ELECTRS_BULK_INDEX_THREADS=2
      - ELECTRS_TXID_LIMIT=200

```

**System running electrs**
Docker, built from the provided Dockerfile

Kixunil:
Is it possible that you terminate electrs with SIGKILL instead of SIGINT?

romanz:
```
2021-10-18T16:14:20.032Z INFO  electrs::db] "./db/signet": 0 SST files, 0 GB, 0 Grows
2021-10-18T16:14:20.033Z DEBUG electrs::db] DB None
```
It seems that the RocksDB directory is empty on startup... 
Could you please check if `./db/signet` is non-empty after the sync is over?

You can also check the following Prometheus metric to see the current DB size:
```
$ curl -s localhost:34224 | grep db_
# HELP index_db_size Index DB size (bytes)
# TYPE index_db_size gauge
index_db_size{type="total"} 25992505
```

romanz:
Cool :)

[...]

PR #579, opened 2021-10-20
Add a note about Dockerfile changes

romanz:
Following https://github.com/romanz/electrs/issues/573#issuecomment-947399630


schildbach:
So you require the user to not use the defaults? I wonder if this case should be caught at runtime, as it seems to be a common "mistake" to me. Especially for people used to the "convention over configuration" mantra.

But yes, for now the sentence lgtm.

Issue #539, opened 2021-10-08
0.9.0 Performance on rPi


Transisto:
is 250 address lookup per minute normal?

public electrumX and blockstream esplora server takes about 3 seconds to lookup 1600 addresses vs 6.5 minutes for electrs.

So Electrs on a rPi is about 120 times slower than with public servers.

Transisto:
![0 9 0](https://user-images.githubusercontent.com/388716/136496705-1d452e89-f0a5-43a9-b0de-179e11471a8d.gif)

romanz:
Could you please reproduce the slowness issue on a testnet wallet (so you could share the xpub here) and attach a DEBUG log?

romanz:
For example, you can use `all all all all all all all all all all all all` mnemonic with `m/44'/1'/0'` derivation - it results in the following testnet master key:
`tpubDDKn3FtHc74CaRrRbi1WFdJNaaenZkDWqq9NsEhcafnDZ4VuKeuLG2aKHm5SuwuLgAhRkkfHqcCxpnVNSrs5kJYZXwa6Ud431VnevzzzK3U`
It currently has 73 transactions (see attached JSON).
[electrum-history.json.gz](https://github.com/romanz/electrs/files/7309289/electrum-history.json.gz)

romanz:
@Transisto BTW, could you please compare `0.8.*` performance to `0.9.0` one?

Transisto:
> @Transisto BTW, could you please compare `0.8.*` performance to the `0.9.0` on your device?

I've tested on somewhat same hardware but 0.8 on umbrel instead of 0.9.0 on mynode,
![umbrel tor](https://user-images.githubusercontent.com/388716/136714982-79e26366-58d2-4263-9810-fae2020025e3.gif)

The connection is over Tor so maybe that's why it can't go all the way to 1600 address without restarting, 
But overall it's about 6x faster.  It did 1250 address lookup in 1min instead of 250

Transisto:
I updated umbrel to 4.4 with electrs 0.9.0 and the speed became the exact same as on the 0.9.0 on mynode.

6 times slower than 0.8.11

romanz:
Could you maybe test `xpub6BiVtCpG9fQPxnPmHXG8PhtzQdWC2Su4qWu6XW9tpWFYhxydCLJGrWBJZ5H6qTAHdPQ7pQhtpjiYZVZARo14qHiay2fvrX996oEP42u8wZy`?
[...]

PR #610, opened 2021-11-07
Allow skipping merkle proofs' during subscription

romanz:
We still support them (by computing them on-demand).

https://github.com/romanz/electrs/commit/3ce23ae01e44ed6183fa95d892953e70a22c8c88#r59562642


romanz:
Tested using [all12_subs.txt](https://github.com/romanz/electrs/files/7491877/_all12_subs.txt) (warm cache).

With `--skip-merkle-proofs-on-subscription`:
```
$ curl -s localhost:4224 | grep 'rpc_duration_sum{method="blockchain.scripthash.subscribe:multi"}'
rpc_duration_sum{method="blockchain.scripthash.subscribe:multi"} 7.838893273

```

Without:
```
$ curl -s localhost:4224 | grep 'rpc_duration_sum{method="blockchain.scripthash.subscribe:multi"}'
rpc_duration_sum{method="blockchain.scripthash.subscribe:multi"} 8.590418878
```

romanz:
The last commit results in a significant performance gain:
```bash
# Without --skip-merkle-proofs-on-subscription:
$ curl -s localhost:4224 | grep 'rpc_duration_sum{method="blockchain.scripthash.subscribe:multi"}'
rpc_duration_sum{method="blockchain.scripthash.subscribe:multi"} 8.89051942

# With --skip-merkle-proofs-on-subscription:
$ curl -s localhost:4224 | grep 'rpc_duration_sum{method="blockchain.scripthash.subscribe:multi"}'
rpc_duration_sum{method="blockchain.scripthash.subscribe:multi"} 4.316420194
```

romanz:
Can be relevant for #539 :)
CC: @transisto

romanz:
The cool thing is that merkle proofs are being client-side cached, so there actually no need to cache them in electrs :)

romanz:
It seems that we get now comparable performance than `0.8.x` branch: https://github.com/romanz/electrs/issues/539#issuecomment-940360858 :eyes:

src/status.rs
```diff
@@ -516,17 +518,48 @@ fn compute_status_hash(history: &[HistoryEntry]) -> Option<StatusHash> {
...
+}
+
 fn filter_block_txs<T: Send>(
     block: Block,
-    f: impl Fn(usize, Transaction, &[Txid]) -> Option<T> + Sync,
-) -> Vec<T> {
-    let txids: Vec<Txid> = block.txdata.par_iter().map(|tx| tx.txid()).collect();
+    map_fn: impl Fn(&Transaction) -> Vec<T> + Sync,
+    cache_proofs: bool,
+) -> impl Iterator<Item = FilteredTx<T>> {
```
Kixunil, line 533: Is there some reason this can't return `impl ParallelIterator`?

Kixunil (approved):
That's some great news! The code looks good too.

src/status.rs
romanz: IIUC, we can't use `for_each()` [here](https://github.com/romanz/electrs/pull/610/files#diff-0bfaab7c77d86eb41c2fe32c5e5df509fa02ab04c97b0ce3216911afaae0d67cR349), since it requires sharing mutable references to `block_entries` and `outpoints` between multiple threads.
Please let me know if there is a better way to do it :)

romanz:
> The cool thing is that merkle proofs are being client-side cached (at least in Electrum and Sparrow), so there actually no need to cache them in electrs :)

I've [flipped the logic](https://github.com/romanz/electrs/pull/610/files#diff-358dacd27838bf24cbb58b091dbdb17fd4e137cc4db957a46909e4a4159e6406R109) so we don't cache merkle proofs by default :)

src/status.rs
Kixunil: May be worth benchmarking and comparing with [lock-free map](https://docs.rs/lockfree/0.5.1/lockfree/map/struct.Map.html)

Or maybe it's possible to somehow collect the generated items with a few more combinators?

romanz: > May be worth benchmarking and comparing with [lock-free map](https://docs.rs/lockfree/0.5.1/lockfree/map/struct.Map.html).

TIL, will try :)

> Or maybe it's possible to somehow collect the generated items with a few more combinators?

Sorry, couldn't find the right way for doing it here:
https://docs.rs/rayon/1.5.1/rayon/iter/trait.ParallelIterator.html

Kixunil:
I wonder is there actually any reason to enable caching? Maybe having the option is not actually useful?

romanz:
> Maybe having the option is not actually useful?

image

PR #612, opened 2021-11-07
Add possibility to print out the program version without actually running it

seberm:
This pull request adds a possiblity to print out the current program version and exit immediately.

There are cases where I need to grab versions of multiple apps on my system. It would be great if I could get electrs version without parsing its logs (`Starting electrs 0.9.2 on x86_64 linux with Config { network: Bitcoin...`).

This is for example also handy when I update nix packages on my `nix-bitcoin` node and I need to recheck the program versions before actually switching to them (`nixos-rebuild switch`).

### DIFF
--- a/internal/config_specification.toml
+++ b/internal/config_specification.toml
@@ -108,6 +108,10 @@ doc = "Exit after the initial sync is over (don't start Electrum server)."
 name = "cache_merkle_proofs"
 doc = "Cache merkle proofs on subscription (if disabled, they are computed on demand)."
 
+[[switch]]
+name = "version"
+doc = "Print out the program version."
+
 [[param]]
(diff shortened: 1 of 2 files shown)


romanz (approved):
Looks good % a small formatting issue:
https://github.com/romanz/electrs/runs/4132455003?check_suite_focus=true#step:4:15

seberm:
The format issue should be fixed. I also changed the `eprintln` to `println`. I think the version should be printed to stdout instead of stderr in this case.

romanz:
Thanks!

Kixunil:
Hmm, I wanted to build it into `configure_me` but this looks simple enough. It'll be easy to replace at any time anyway.

PR #615, opened 2021-11-08
Remove `verbose` configuration

romanz:
Please use `RUST_LOG` to specify logging verbosity.
See https://github.com/env-logger-rs/env_logger/ for details.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -287,12 +287,10 @@ impl Config {
             }
         });
 
-        let level = match config.verbose {
-            0 => log::LevelFilter::Error,
-            1 => log::LevelFilter::Warn,
-            2 => log::LevelFilter::Info,
-            _ => log::LevelFilter::Debug,
-        };
+        if config.verbose > 0 {
+            eprintln!("Error: please use `log_filter` to set logging verbosity",);
+            std::process::exit(1);
+        }
 
         let index_lookup_limit = match config.index_lookup_limit {
             0 => None,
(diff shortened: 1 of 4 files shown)


Kixunil:
Why? I would really, really like to set logging in configuration file. :(

romanz:
You're right, let's replace it with a new configuration - to be used to specify filters via https://docs.rs/env_logger/0.9.0/env_logger/struct.Builder.html#method.parse_filters.
WDYT?

Kixunil:
Yeah, additional config option called `log_filters` sounds good.

Kixunil:
One more thing: it leaks internal structure of modules in the configuration by allowing to set filters per-module. I think it's fine but we should document that it's forever unstable and people should only set specific modules temporarily when we tell them during debugging.

src/config.rs
```diff
@@ -287,12 +287,11 @@ impl Config {
...
 
-        let level = match config.verbose {
-            0 => log::LevelFilter::Error,
-            1 => log::LevelFilter::Warn,
-            2 => log::LevelFilter::Info,
-            _ => log::LevelFilter::Debug,
-        };
+        if config.verbose > 0 {
+            eprintln!("Error: please use `log_filters` to set logging verbosity",);
+            std::process::exit(1);
```
Kixunil, line 292: Shouldn't release notes say "removed" instead of "deprecated"?

romanz: Updated the PR and the relevant commits.

PR #606, opened 2021-11-01
Creating the quickstart guide

k3tan172:
Electrs is an incredibly critical piece of software and requires a quick and easy way to get up and running. I have written this guide for beginners hoping others can use it. Feedback welcome.

PR #618, opened 2021-11-12
Fixup `--log-filters` documentation and examples

romanz:
Drop `-v` and `--verbose` mentions.

### DIFF
--- a/doc/upgrading.md
+++ b/doc/upgrading.md
@@ -2,6 +2,16 @@
 
 * If you use `verbose` (or `-v` argument), switch to `log_filters` (or `RUST_LOG` environment variable).
   Please note that it allows to set per-module filters, but module naming is considered unstable.
+  If you have used `-vv` (the value suggested in the documentation), switch to `--log-filters INFO`:
+
+
+|Log filter|Old `verbose` value|Description                                                           |
+|----------|-------------------|----------------------------------------------------------------------|
+|ERROR     |                  0|Only fatal errors                                                     |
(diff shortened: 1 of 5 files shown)


romanz:
Following https://github.com/romanz/electrs/pull/606#issuecomment-966982237

romanz:
cc: @Kixunil @k3tan172

doc/upgrading.md
```diff
@@ -2,6 +2,7 @@
 
 * If you use `verbose` (or `-v` argument), switch to `log_filters` (or `RUST_LOG` environment variable).
   Please note that it allows to set per-module filters, but module naming is considered unstable.
+  If you have used `-vv` (the default setting), switch to `--log-filters INFO`.
```
Kixunil, line 5: I think this should say "the value suggested in the documentation" instead of "the default setting" since the default was ERROR.

Kixunil (changes requested):
I think it'd be good to add a table like this somewhere:

|Log level|Old verbose value|                                                                     Description|
|------------|------------------------|-------------------------------------------------------------------------|
| ERROR|                           0|                                                              Only fatal errors|
|   WARN|                           1|                   Things that could indicate serious problems|
|     INFO|                           2|                     Various significant events and suggestions|
| DEBUG|                           3|Details that could be useful when debugging - only use when debugging!|
|  TRACE|                          4|**Very** detailed information - only use when debugging!|

Kixunil (approved):
Looks good.

PR #620, opened 2021-11-13
Add dependabot config for autmatic cargo update PRs

dbast:
Merging this opens update PRs like they can be seen here https://github.com/dbast/electrs/pulls

Proposed updates can be merged if CI is green or closed to be ignored.

There are also commands to tell dependabot e.g. to ignore specific major versions `@dependabot ignore this major version` as described in the PRs.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,6 @@
+version: 2
+updates:
+  - package-ecosystem: "cargo"
+    directory: "/"
+    schedule:
+      interval: "weekly"


Kixunil:
Is there a way to set MSRV?

romanz (approved):
Cool, many thanks!

romanz:
> Is there a way to set MSRV?

Not sure, looking [here](https://docs.github.com/en/code-security/supply-chain-security/keeping-your-dependencies-updated-automatically/configuration-options-for-dependency-updates) seems that there is no explicit configuration - but I may be wrong.

PR #624, opened 2021-11-14
Dockerfile: re-add curl, so it can be used for docker health checks

schildbach:
Note: I've decided to not add a default health check to the Dockerfile, since the monitoring port is different for each network. I could add a default for mainnet, but then by default testnet and signet installations would be unhealthy.

If you want to add a health check to your docker-compose.yaml, you can use:

```
    healthcheck:
      test: ["CMD-SHELL", "curl -fSs http://localhost:4224/ || exit 1"]
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -4,7 +4,7 @@
 
 FROM debian:bullseye-slim as base
 RUN apt-get update -qqy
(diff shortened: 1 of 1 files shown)


romanz (approved):
Thanks!

Issue #619, opened 2021-11-12
Prepare an "asciinema" screencast for a quickstart installation


romanz:
> Maybe we can use https://asciinema.org/ to prepare a short screencast that shows how to install and run electrs on a Linux machine? This way, beginner users will also have the commands' output available (which can help with troubleshooting).

_Originally posted by @romanz in https://github.com/romanz/electrs/issues/606#issuecomment-956156699_

romanz:
@Kixunil @k3tan172 WDYT about https://asciinema.org/a/jrh0CqCnImvtIOnW7N4uachkR?

romanz:
TIL that you can install https://asciinema.org client using `sudo apt install asciinema` :)

Kixunil:
Could the waiting parts be sped up?

romanz:
The `speed` option allows running it faster:
https://asciinema.org/a/jrh0CqCnImvtIOnW7N4uachkR?rows=30&speed=3

romanz:
Manually starting electrs sync:
https://asciinema.org/a/kGPYuL75Ww3hi0pw3DOziv0DT?speed=3&cols=200

PR #625, opened 2021-11-16
Add asciinema screencasts for quickstart installation and usage

romanz:
Following #619

### DIFF
--- a/doc/install.md
+++ b/doc/install.md
@@ -1,3 +1,21 @@
+## Quickstart
+
+<details>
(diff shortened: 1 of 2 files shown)


romanz:
CC: @k3tan172 @Kixunil

Kixunil:
I think we should somehow briefly document when the quickstarts are applicable so that people know whether the assumptions about their system are correct.

romanz:
Added a short `details` section before each screencast - WDYT?

Kixunil (approved):
Looks good, we'll see if it helps after merging I guess...

PR #627, opened 2021-11-17
contrib: rename xpub into get_balance

jroulin:
- support a single address or an xpub address
- detail the balance per address when using an xpub


romanz (approved):
Thanks!

PR #634, opened 2021-12-01
Warn on attempt to connect over SSL

cbergqvist:
Usecase: I was changing my client & server configuration around a lot (trying raw Bitcoin Core, turning on/off Tor etc) and ended up trying to connect with SSL activated on the client but without running Electrs behind Nginx. Naturally it failed, but to save my future self or the next person some time, I suggest adding a clear warning in the log.

Feel free to tidy up the change or instruct me in how to best improve it. My Rust-fu is weak.

### DIFF
--- a/src/server.rs
+++ b/src/server.rs
@@ -219,11 +219,18 @@ fn recv_loop(peer_id: usize, stream: &TcpStream, server_tx: Sender<Event>) -> Re
     let msg = Message::New(stream.try_clone()?);
     server_tx.send(Event { peer_id, msg })?;
 
+    let mut first_line = true;
     for line in BufReader::new(stream).lines() {
+        if let Err(e) = &line {
+            if first_line && e.kind() == std::io::ErrorKind::InvalidData {
+                warn!("InvalidData on first line may indicate client attempted to connect using SSL when server expects unencrypted communication.")
+            }
+        }
         let line = line.with_context(|| format!("{}: recv failed", peer_id))?;
         debug!("{}: recv {}", peer_id, line);
         let msg = Message::Request(line);
         server_tx.send(Event { peer_id, msg })?;
+        first_line = false;
     }
 
     debug!("{}: disconnected", peer_id);
--- a/src/thread.rs
+++ b/src/thread.rs
@@ -9,6 +9,7 @@ where
         .spawn(move || {
             if let Err(e) = f() {
                 warn!("{} thread failed: {}", name, e);
+                e.chain().skip(1).for_each(|e| warn!("because: {}", e));
             }
         })
         .expect("failed to spawn a thread")


src/server.rs
```diff
@@ -219,11 +219,19 @@ fn recv_loop(peer_id: usize, stream: &TcpStream, server_tx: Sender<Event>) -> Re
     let msg = Message::New(stream.try_clone()?);
     server_tx.send(Event { peer_id, msg })?;
 
+    let mut first_line = true;
     for line in BufReader::new(stream).lines() {
+        if let Err(ref e) = line {
+            warn!("{}: recv error: {:?} - {}", peer_id, e.kind(), e);
```
Kixunil, line 225: Isn't this logged already?

src/server.rs
```diff
@@ -219,11 +219,19 @@ fn recv_loop(peer_id: usize, stream: &TcpStream, server_tx: Sender<Event>) -> Re
     let msg = Message::New(stream.try_clone()?);
     server_tx.send(Event { peer_id, msg })?;
 
+    let mut first_line = true;
     for line in BufReader::new(stream).lines() {
+        if let Err(ref e) = line {
```
Kixunil, line 224: Nit: I believe `if let Err(error) = &line {` is idiomatic these days.

src/server.rs
```diff
@@ -219,11 +219,19 @@ fn recv_loop(peer_id: usize, stream: &TcpStream, server_tx: Sender<Event>) -> Re
...
+            warn!("{}: recv error: {:?} - {}", peer_id, e.kind(), e);
+            if first_line && e.kind() == std::io::ErrorKind::InvalidData {
+                warn!("InvalidData on first line may indicate client attempted to connect using SSL when server expects unencrypted communication.")
+            }
+        }
         let line = line.with_context(|| format!("{}: recv failed", peer_id))?;
```
Kixunil, line 230: Was thinking of something like this, not sure if it's better:

```rust
let line = line.map_err(|error| {
    // warn!() if it's not redundant
    if first_line && error.kind() == std::io::ErrorKind::InvalidData {
        warn!("InvalidData on first line may indicate client attempted to connect using SSL when server expects unencrypted communication.")
    }
    first_line = false;
    error.context(format!("{}: recv failed", peer_id))
})?
```

Kixunil (commented):
Concept ACK.

src/server.rs
cbergqvist: The only thing emitted when connecting with an SSL client without this change is:
```
 [2021-12-01T20:40:47.681Z WARN  electrs::thread] recv_loop thread failed: 2: recv failed
```
Knowing the error kind and string can be helpful.

src/server.rs
cbergqvist: I tried doing that but `error.context()` doesn't exist, would need the `Result`.

cbergqvist: I avoided the extra `warn!` unless it's not an `InvalidData` error in the newest version.

Kixunil: Ah, didn't realize `error` is not `anyhow::Error`.

src/server.rs
Kixunil: Hmm, would be a bit nicer to fix error reporting done by a caller.

cbergqvist: Yeah, I agree that having `spawn()` in *thread.rs* unwrap the `anyhow` error and append information about the inner error. I'll see if I can figure out how to do that.

cbergqvist: Alright, 844d7c2 results in the following output:
```
[2021-12-01T21:58:36.290Z WARN  electrs::server] InvalidData on first line may indicate client attempted to connect using SSL when server expects unencrypted communication.
[2021-12-01T21:58:36.290Z WARN  electrs::thread] recv_loop thread failed: 0: recv failed
[2021-12-01T21:58:36.290Z WARN  electrs::thread] because: stream did not contain valid UTF-8
```

Kixunil (approved):
Looks good!

It could happen that two threads error at the same time and mix messages but hopefully it's unlikely and I really like them separated.

romanz (approved):
Many thanks!

PR #653, opened 2022-01-08
Fix `txid` index collision handling

romanz:
Otherwise, we may fail to retrieve existing transaction whose `txid[..8]` collides with other transactions.


romanz:
Following https://github.com/romanz/electrs/discussions/651#discussioncomment-1930655 :)

PR #658, opened 2022-01-10
Clearly state that altcoins are not supported.

Kixunil:
I think we wanted to add this but forgot. Used the word "altcoin" instead of "shitcoin" to avoid "x is not a shitcoin but an altcoin" arguments.


romanz:
Thanks :)

PR #674, opened 2022-01-26
More accurate free space required for reindexing

brandonblack:
The free space required peaks higher than the final size before and during compaction.

Final size is ~40G, peak was ~60G on my machine.


romanz:
LGTM. thanks!
@Kixunil WDYT?

Kixunil:
Yeah, makes sense, completely forgot about the spike when I wrote it! (I'm assuming the numbers are correct, didn't test.)

PR #686, opened 2022-02-17
Use configure_me to skip default config files

### LINKED ISSUE #217: Unnecessary access of electrs homedir, prevents systemd hardening
When running Electrs as a systemd unit, I usually harden the configuration by including the following access limitation:

```sh
# Deny access to /home, /root and /run/user
ProtectHome=true
```

In my specific configuration, the user `electrs` does not need to have a home directory, as the configuration file is located at `/etc/electrs/electrs.conf` and the `db-dir` is on a different drive. But if I include the restriction above, Electrs fails to start with the following error message:

```sh
Error: Failed to read configuration file /home/electrs/.electrs/config.toml: Permission denied (os error 13)
```

This seems to indicate that Electrs tries to access its service user's home directory nonetheless, and fails because it's not accessible. I think it would be good practice to not access this directory if not necessary.

Here's the full systemd unit file for reference:

```systemd
[Unit]
Description=Electrs server daemon
After=bitcoind.service

[Service]

# Service execution
###################

ExecStart=/usr/bin/electrs --conf /etc/electrs/electrs.conf

# Process management
####################

Type=simple
Restart=always
TimeoutSec=120
RestartSec=30
KillMode=process


# Directory creation and permissions
####################################

# Run as electrs:bitcoin
User=electrs
Group=bitcoin

# /run/electrs
RuntimeDirectory=electrs
RuntimeDirectoryMode=0710

# Hardening measures
####################

# Provide a private /tmp and /var/tmp.
PrivateTmp=true

# Mount /usr, /boot/ and /etc read-only for the process.
ProtectSystem=full

# Deny access to /home, /root and /run/user
ProtectHome=true

# Disallow the process and all of its children to gain
# new privileges through execve().
NoNewPrivileges=true

# Use a new /dev namespace only populated with API pseudo devices
# such as /dev/null, /dev/zero and /dev/random.
PrivateDevices=true

# Deny the creation of writable and executable memory mappings.
MemoryDenyWriteExecute=true

[Install]
WantedBy=multi-user.target
```

Kixunil:
This removes the feature for skipping default config files and replaces
it with a configure_me implementation.

Closes #217

~~Note: `cfg_me` supporting this will be released soon.~~ Done.
Sorry for this taking so long time.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -235,9 +235,9 @@ dependencies = [
 
 [[package]]
 name = "configure_me_codegen"
-version = "0.4.1"
+version = "0.4.2"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "a7ddbd860cbff075ceadf4714922a73afbe8bf53d9ed9add6c6b733feeb6f3c7"
+checksum = "9aa65df44cc55922b4e22f49939727156a8de25492bf6bd2462a75afc5e155d6"
 dependencies = [
  "cargo_toml",
  "fmt2io",
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -16,7 +16,6 @@ build = "build.rs"
 default = ["metrics"]
 metrics = ["prometheus", "tiny_http"]
 metrics_process = ["prometheus/process"]
-ignore_default_config_files = []
 
 [package.metadata.configure_me]
 spec = "internal/config_specification.toml"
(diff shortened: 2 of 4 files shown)


romanz:
Many thanks!

bitagilis:
Can you write some documentation on how to enable such feature? Does it require to pass some command line argument?

I tried with the following command line arguments: --features "ignore_default_config_files" and '--skip_default_conf_files' without any success as electrs keeps failing to start indicating that such arguments are unknown.

Thanks.

Issue #623, opened 2021-11-13
incompatibility with the new version of rocksdb


prusnak:
Somwhere between 6.23.3 and 6.25.3 the API or rocksdb has changed:

```
   Compiling electrs-rocksdb v0.15.0-e2
error[E0308]: mismatched types
   --> /private/tmp/nix-build-electrs-0.9.2.drv-0/electrs-0.9.2-vendor.tar.gz/electrs-rocksdb/src/db_options.rs:477:56
    |
477 |                 ffi::rocksdb_filterpolicy_create_bloom(bits_per_key)
    |                                                        ^^^^^^^^^^^^ expected `f64`, found `i32`
    |
help: you can convert an `i32` to an `f64`, producing the floating point representation of the integer
    |
477 |                 ffi::rocksdb_filterpolicy_create_bloom(bits_per_key.into())
    |                                                        ^^^^^^^^^^^^^^^^^^^

error[E0308]: mismatched types
   --> /private/tmp/nix-build-electrs-0.9.2.drv-0/electrs-0.9.2-vendor.tar.gz/electrs-rocksdb/src/db_options.rs:479:61
    |
479 |                 ffi::rocksdb_filterpolicy_create_bloom_full(bits_per_key)
    |                                                             ^^^^^^^^^^^^ expected `f64`, found `i32`
    |
help: you can convert an `i32` to an `f64`, producing the floating point representation of the integer
    |
479 |                 ffi::rocksdb_filterpolicy_create_bloom_full(bits_per_key.into())
    |                                                             ^^^^^^^^^^^^^^^^^^^

```

prusnak:
It seems the fractional bits_per_key was introduced quite some time ago, see https://github.com/facebook/rocksdb/commit/57f30322855b6ed38296ec0f4c6fb03b095fe0b3 (27 Nov 2019)

romanz:
Many thanks for reporting this issue!

Due to multiple compatibility issues (mostly with ARM platforms - e.g. #403), we have forked the `rust-rocksdb` crate (https://crates.io/crates/electrs-rocksdb), in order to "pin" the C++ RocksDB to 6.11.4 release (#473) to allow dynamic linking on Debian 11.

So indeed, it probably won't work with the latest RocksDB releases :(

romanz:
FWIW, we have https://github.com/romanz/electrs/issues/201 :)

Kixunil:
[...]

PR #696, opened 2022-03-17
Add guide for other Ubuntu & Debian releases to compile and install librocksdb

ayanamidev:
Current docs lacks instruction to install electrs and librocksdb for other releases of Ubuntu and Debian which includes LTS versions so adding it.

cc @romanz

### DIFF
--- a/doc/install.md
+++ b/doc/install.md
@@ -59,6 +59,15 @@ On [Debian 11 (bullseye)](https://packages.debian.org/bullseye/librocksdb-dev) a
 $ sudo apt install librocksdb-dev=6.11.4-3
 ```
 
+For other versions of Debian or Ubuntu, you can build librocksdb and install inside `/usr/local` directory using following command.
+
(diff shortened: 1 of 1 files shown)


Kixunil:
Hmm? Why would one manually build rocksdb if static linking achieves the same?

ayanamidev:
@Kixunil With my understanding, shouldn't electrs be compiled with the specific version of rocksdb? #623

Kixunil:
There are two ways to compile electrs: static linking and dynamic. If you use static linking, you don't need to mess with rocksdb library manually at all. Dynamic linking requires correct version but you get the benefits of dynamic linking only if you use system library. Manually compiling it may end up being worse. E.g. your command doesn't verify authenticity of the library, so you could get a backdoored version.

Long term it'd be great to figure out a way to support multiple versions but meanwhile it may be better to recommend static linking.

PR #740, opened 2022-07-03
Replace RocksDB with sled

KhaledEmaraDev:


Kixunil:
AFAIK `sled` is worse than rocksdb, so it should be compile-time selectable, ideally using Clean Architecture.

KhaledEmaraDev:
@Kixunil I'm working on making both implement a trait. Should I convert this PR to a draft one?

Kixunil:
@KhaledEmaraDev I think it's a good idea

romanz:
@KhaledEmaraDev Thanks for opening this PR!
Could you please adapt it to support both DBs, and attach the sled performance metrics?

KhaledEmaraDev:
@romanz working on it.

romanz:
We should also consider https://github.com/cberner/redb, following https://github.com/romanz/electrs/issues/753.

KhaledEmaraDev:
@romanz Will work on including it.

casey:
btw, the author of redb just added benchmarks vs rocksdb, and it looks like it's way faster: https://github.com/cberner/redb#benchmarks

Kixunil:
Wow, that looks great! Add some RPC optimizations and we may finally fix perf issues.

KhaledEmaraDev:
This is replaced by #765.

PR #762, opened 2022-08-16
WIP: Feature/custom signet

Sosthene00:
This PR attempts to solve the issue raised in #418 
It is pretty straightforward, a `signet_magic` can be added as an option. If it is not set the magic is the default value for the network (`network.magic()`).
It is not fully operational though, I tried to run it against a custom signet and it seems that it downloads some blocks then fails for some reason. Here's the full log of a typical run:
```
Starting electrs 0.9.8 on x86_64 linux with Config { network: Signet, db_path: "/home/electrs/db/signet", daemon_dir: "/data/signet", daemon_auth: CookieFile("/data/signet/.cookie"), daemon_rpc_addr: 192.168.240.3:38332, daemon_p2p_addr: 192.168.240.3:38333, electrum_rpc_addr: 127.0.0.1:60601, monitoring_addr: 127.0.0.1:34224, wait_duration: 10s, jsonrpc_timeout: 15s, index_batch_size: 10, index_lookup_limit: None, reindex_last_blocks: 0, auto_reindex: true, ignore_mempool: false, sync_once: false, disable_electrum_rpc: false, server_banner: "Welcome to electrs 0.9.8 (Electrum Rust Server)!", signet_magic: 3278571876, args: [] }
[2022-08-16T13:50:01.595Z INFO  electrs::metrics::metrics_impl] serving Prometheus metrics on 127.0.0.1:34224
[2022-08-16T13:50:01.595Z INFO  electrs::server] serving Electrum RPC on 127.0.0.1:60601
[2022-08-16T13:50:01.601Z DEBUG tiny_http] Server listening on 127.0.0.1:34224
[2022-08-16T13:50:01.606Z DEBUG tiny_http] Running accept thread
[2022-08-16T13:50:01.670Z INFO  electrs::db] "/home/electrs/db/signet": 13 SST files, 0.000010465 GB, 0.000000013 Grows
[2022-08-16T13:50:01.670Z DEBUG electrs::db] DB Some(Config { compacted: false, format: 0 })
[2022-08-16T13:50:01.688Z DEBUG bitcoincore_rpc] JSON-RPC request: getblockchaininfo []
[2022-08-16T13:50:01.689Z DEBUG bitcoincore_rpc] JSON-RPC error for getblockchaininfo: RpcError { code: -28, message: "Loading wallet…", data: None }
[2022-08-16T13:50:01.689Z INFO  electrs::daemon] waiting for RPC warmup: Loading wallet…
[...]

PR #768, opened 2022-09-05
Takes signet magic in hex as in bitcoin core logs

Sosthene00:
This is a follow-up on #762, to allow the `signet-magic` option to be passed as the hex string that is found in Bitcoin Core logs.

Previously the magic was passed directly as a u32, which was cumbersome as user had to manually reverse the bytes in the magic given by Core and convert it to an int.

With this PR, you can grep `magic` in Bitcoin Core logs (let's say it's `deadbeef`), and then directly pass it as argument `--signet-magic deadbeef`. Previously you would have to first figure out that it corresponds to `4022250974`.

I should have directly made it like that sorry.

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -234,7 +234,7 @@ impl Config {
 
         let magic = match (config.network, config.signet_magic) {
             (Network::Signet, Some(magic)) => u32::from_str_radix(&magic, 16)
-                .expect("Invalid hex string")
+                .expect("Invalid signet hex magic")
                 .swap_bytes(),
             (network, None) => network.magic(),
             (_, Some(_)) => {
--- a/src/config.rs
+++ b/src/config.rs
@@ -233,7 +233,9 @@ impl Config {
         };
 
         let magic = match (config.network, config.signet_magic) {
-            (Network::Signet, Some(magic)) => magic,
+            (Network::Signet, Some(magic)) => u32::from_str_radix(&magic, 16)
+                .expect("Invalid hex string")
+                .swap_bytes(),
             (network, None) => network.magic(),
             (_, Some(_)) => {
                 eprintln!("Error: signet magic only available on signet");
--- a/internal/config_specification.toml
+++ b/internal/config_specification.toml
@@ -134,5 +134,5 @@ doc = "Logging filters, overriding `RUST_LOG` environment variable (see https://
 
 [[param]]
 name = "signet_magic"
-type = "u32"
-doc = "network magic for custom signet network (signet only)"
+type = "String"
+doc = "network magic for custom signet network in hex format, as found in Bitcoin Core logs (signet only)"


src/config.rs
```diff
@@ -233,7 +233,9 @@ impl Config {
         };
 
         let magic = match (config.network, config.signet_magic) {
-            (Network::Signet, Some(magic)) => magic,
+            (Network::Signet, Some(magic)) => u32::from_str_radix(&magic, 16)
+                .expect("Invalid hex string")
```
romanz, line 237:
```suggestion
                .expect("Invalid signet hex magic")
```

romanz (approved):

romanz:
Many thanks!

src/config.rs
```diff
@@ -233,7 +233,9 @@ impl Config {
         };
 
         let magic = match (config.network, config.signet_magic) {
-            (Network::Signet, Some(magic)) => magic,
+            (Network::Signet, Some(magic)) => u32::from_str_radix(&magic, 16)
+                .expect("Invalid signet hex magic")
```
Kixunil, line 237: It'd be much better to have a user-friendly error message:

```rust
.unwrap_or_else(|error| {
    eprintln!("Error: the signet magic '{}' is not a valid hex string: {}", magic, error);
    std::process::exit(1);
})
```

src/config.rs
```diff
@@ -233,7 +233,9 @@ impl Config {
...
+                .swap_bytes(),
```
Kixunil, line 238: Shouldn't this be `to_be` or `to_le`? I don't know the details of network magic and Core serialization but it does look suspicious.

Kixunil (commented):
Sorry for late review, this looks suspicious.

src/config.rs
Sosthene00: Hi, it will probably clearer this way in case user messes up, thanks

src/config.rs
Sosthene00: I'm no specialist either, but I made a little research about how people usually handle this kind of situations in Rust and this seemed the most straightforward way. I'll take a look again and try to make a better case for it (or switch to `to_le` if it's actually better).

Kixunil: https://docs.rs/bitcoin/latest/bitcoin/network/constants/enum.Network.html#method.magic says it's encoded as little endian, so I think `u32::from_le` should be correct. Anyway, the API should be [improved on `rust-bitcoin` side](https://github.com/rust-bitcoin/rust-bitcoin/issues/1266).

Kixunil: Sure, just note that the new version of rust-bitcoin wasn't released yet.

Sosthene00: Haha, yeah I know, just giving myself a little head start on this one

PR #786, opened 2022-11-01
Don't panic in case of an invalid block header height

### LINKED ISSUE #784: Bug: crash on invalid headers params
Hi

https://github.com/romanz/electrs/blob/a3ef0fa895285dc2fb7672c02080b49993f8880a/src/electrum.rs#L230

The .unwrap() there causes electrs to crash if you request a block header that doesn't exist.

romanz:
Fix #784


romanz:
CC: @benma

PR #801, opened 2022-11-14
cross-compilation instructions using debootstrap

zciendor:
Added instructions how to cross-compile for a different architecture and OS version (Debian based) than the target system.

### DIFF
--- a/doc/install.md
+++ b/doc/install.md
@@ -89,13 +89,13 @@ $ sudo apt update
 If you use `cargo` from the repository
 
 ```bash
-$ sudo apt install gcc-aarch64-linux-gnu gcc-aarch64-linux-gnu libc6-dev:arm64 libstd-rust-dev:arm64
+$ sudo apt install gcc-aarch64-linux-gnu g++-aarch64-linux-gnu libc6-dev:arm64 libstd-rust-dev:arm64
 ```
 
 If you use Rustup:
 
 ```bash
-$ sudo apt install gcc-aarch64-linux-gnu gcc-aarch64-linux-gnu libc6-dev:arm64
+$ sudo apt install gcc-aarch64-linux-gnu g++-aarch64-linux-gnu libc6-dev:arm64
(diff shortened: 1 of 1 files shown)


doc/install.md
```diff
@@ -89,13 +89,13 @@ $ sudo apt update
 If you use `cargo` from the repository
 
 ```bash
-$ sudo apt install gcc-aarch64-linux-gnu g++-aarch64-linux-gnu libc6-dev:arm64 libstd-rust-dev:arm64
```
romanz, line 92: Is it needed (here and below)?

romanz (changes requested):

doc/install.md
```diff
@@ -105,6 +105,66 @@ If you decided to use the system rocksdb (recommended if the target OS supports
 $ sudo apt install librocksdb-dev:arm64
 ```
 
+#### Preparing for cross compilation on a different (Debian-based) OS distribution/version
+
+If your build system runs on a different OS distribution and/or release than the target system electrs is going to run on, you may run into `GLIBC` version issues like:
```
Kixunil, line 110: Cool! Maybe this should contain a note that unless one runs into libc issue, avoiding debootstrap is faster?

doc/install.md
zciendor: Depends. If you decide for a debootstrap-only approach it's not needed here. It's definitely needed within the debootstrap chroot. But if you build for multiple targets - some for a different os version and some for the same one as the build host - then it's needed in both places.

zciendor: Or what do you mean with "it"? The whole line or just the g++ package?

doc/install.md
zciendor: Done

doc/install.md
zciendor: Actually it's a section of the document I didn't touch. But it's the same typo as in PR #800 so I fixed it here anyway.

romanz:
Thanks!

PR #823, opened 2023-01-07
WIP upgrade with Magic type

Sosthene00:
This is a follow-up of #768 
Since then rust-bitcoin has merged a [PR](https://github.com/rust-bitcoin/rust-bitcoin/pull/1288) that creates a new `Magic` type and modifies the api.
This PR is only a WIP/placeholder while the change is not tagged in a version of rust-bitcoin, its scope is only the necessary changes to make it work with the new way rust-bitcoin handles network's magic.

### DIFF
--- a/src/p2p.rs
+++ b/src/p2p.rs
@@ -7,7 +7,8 @@ use bitcoin::{
     },
     hashes::Hash,
     network::{
-        address, constants,
+        address,
+        constants::{self, Magic},
         message::{self, CommandString, NetworkMessage},
         message_blockdata::{GetHeadersMessage, Inventory},
         message_network,
@@ -130,7 +131,7 @@ impl Connection {
         network: Network,
         address: SocketAddr,
         metrics: &Metrics,
-        magic: bitcoin::network::Magic,
+        magic: Magic,
     ) -> Result<Self> {
         let conn = Arc::new(
             TcpStream::connect(address)
@@ -337,7 +338,7 @@ fn build_version_message() -> NetworkMessage {
 }
 
 struct RawNetworkMessage {
-    magic: bitcoin::network::Magic,
+    magic: Magic,
     cmd: CommandString,
     raw: Vec<u8>,
 }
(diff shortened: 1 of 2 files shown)


src/p2p.rs
```diff
@@ -214,7 +215,7 @@ impl Connection {
             }
             let raw_msg = match raw_msg {
                 Ok(raw_msg) => {
-                    assert_eq!(raw_msg.magic, magic);
+                    assert_eq!(raw_msg.magic.to_hex(), magic.to_hex());
```
Kixunil, line 218: Why? `ToHex` trait will be removed eventually, maybe in the upcoming release.

Sosthene00: `raw_msg.magic` is `[u32;4]`, `magic` is, well, `Magic`, it seemed easier to assert their hex form. I wasn't aware that `ToHex` was going to be deprecated. I'll rewrite that part.

Sosthene00: sorry `u32` I mean

Sosthene00: Pushed a new commit with `assert_eq!(Magic::from_bytes(raw_msg.magic.to_be_bytes()), magic)` instead, I reread the discussion about endianness in the rust-bitcoin PR and I think I got it right ? Alternative would be `assert_eq!(raw_msg.magic, u32::from_be_bytes(magic.to_bytes()))`, and I don't know if one is better than the other.

Kixunil: You should just change the type in `RawNetworkMessage` from `u32` to `Magic` and then it should work out of the box.

Sosthene00: ugh, didn't figure I could do that. Done.

Kixunil (approved):

romanz (approved):
LGTM, thanks!

romanz:
Let's wait for a new https://github.com/rust-bitcoin/rust-bitcoin release :)

romanz:
Please rebase over latest master :)

Sosthene00:
Sorry I just saw your message, thanks

PR #850, opened 2023-02-25
Update tempfile to 3.4.0

romanz:
Following https://github.com/romanz/electrs/security/dependabot/5

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -360,6 +360,27 @@ dependencies = [
 ]
 
 [[package]]
+name = "errno"
+version = "0.2.8"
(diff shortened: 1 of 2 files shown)


Kixunil:
Looks like you just leaked something?

romanz:
Leaked?

romanz:
IIUC, this is an issue in `tempfile` which is used only by unit-tests.

Kixunil:
The link is inaccessible to me, presumably because the reason you upgraded should be secret.

romanz:
:facepalm:

PR #773, opened 2022-10-07
Bump MSRV to 1.63.0 and update dependencies

romanz:
Fixes #753 (allow rocksdb build on Mac OS).
Fixes #877 (support GCC 13).

PR #871, opened 2023-04-09
remove txid table if txindex is enabled

### LINKED ISSUE #798: Feature: no txid table when txindex is enabled

the [txid](https://github.com/romanz/electrs/blob/master/doc/schema.md#transaction-id-index-txid) table takes up space (is there a way to see how much?), and this data is redundant if bitcoind has txindex enabled

if txindex is enabled, this table is removed if it exists (to reclaim storage space when upgrading), and not indexed. getrawtransaction calls dont need a blockhash

none

**Additional context**

useful info for implementing this:

this bitcoin-cli command shows if txindex is enabled or not and gives some info on it. the top level json object will be empty if txindex is not enabled
```
$ bitcoin-cli -testnet getindexinfo txindex
{
  "txindex": {
    "synced": true,
    "best_block_height": 2405719
  }
}
```

antonilol:
fixes #798
depends on #773 for clippy in the ci to succeed (there is a clippy bug in this rust version)

With bitcoind's txindex the txid column family (i called this 'table' in the linked issue) is not necessary and takes up about a quarter of the DB on disk. On testnet my DB was 2.62 GB before this PR, and 1.91 after, on mainnet I expect it to free around 10 GB of the 40 GB currently used.

If txindex is turned on, the txid cf will be emptied and not indexed, if it is turned off, electrs will reindex this cf

todo:

- [x] Working change that checks if txindex is on and empties the txid cf
- [x] Reindex the txid cf if txindex is turned off
- [ ] Resolve todos
- [ ] Test
- [ ] Doc comments if any changes need explanation


antonilol:
i let this rot for too long, will look into it again after #1043

PR #884, opened 2023-05-29
Move 'hex-lit' to [dev-dependencies]

romanz:


Kixunil:
Note that while not wrong, `bitcoin` already depends on it anyway, since there are non-dev constants that are easier to write as hex.

PR #906, opened 2023-07-21
Perf improv

RCasatta:
These are some performance improvements I found by looking at the code after the announcement of the release 0.10.0

I am not sure if they are on the hot path since I didn't find benches or observation data for live electrs instances that would be extremely interesting to see.

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -73,26 +73,28 @@ impl Stats {
     }
 }
 
-struct IndexResult {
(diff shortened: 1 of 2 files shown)


romanz:
Many thanks!
I will push the two last commits in #909 and will rebase https://github.com/romanz/electrs/pull/906/commits/152233b3d7e5256d7e31cb2966cc8d5ecc4cb95c over it.

Issue #904, opened 2023-07-17
Feature:


matthiasdebernardini:
**Is your feature request related to a problem? Please describe.**
Configuring a proxy to add ssl is too much work for something so necessary and rote, none of the proxy solutions are simple enough.

**Describe the solution you'd like**
To have a https mode be a cli option and not have to configure additional software to get basic security.

**Describe alternatives you've considered**
hitch and nginx

**Additional context**
There are crates in rust that help with this, including creating the certs (which I also shouldn't have to do)

romanz:
PRs are welcome :)

matthiasdebernardini:
I think the easiest way to add this would be to use axum and rcgen to spin up another server and do all the wiring inside the binary (so electrs is being tunneled through axum), but I am not sure if you would be happy about that.

romanz:
Would it be possible to hide this feature behind a new [Cargo feature](https://doc.rust-lang.org/cargo/reference/features.html)?

cryptoquick:
This is related to #820, and I was actually about to get started on this. A part of me agrees that axum might be a good approach, since it comes with concurrency and async/await support built in, but I'd want it to be compatible with the Esplora API here:
https://github.com/Blockstream/electrs/blob/adedee15f1fe460398a7045b292604df2161adc0/src/rest.rs

It might just be easier to try to port over the hyper 0.14 implementation and then update it to 1.0. I'm unsure of the concurrency implications of this without doing a deep dive.

I'm also not sure if there's any other Electrum REST API I can reference as I go to implement this, but that'll be my reference for now.

I'd be happy to gate this work behind a cargo feature. If anyone wants to help, let me know, and I can provide access to my electrs fork.

cryptoquick:
Looking through the Esplora code, it looks like they're already using tokio and rolling their own server implementation, so I agree, axum is the better choice here.

cryptoquick:
@romanz Let's say I wanted to get the block tip hash, how might I do it within the approach I've started here?
https://github.com/romanz/electrs/pull/911

In particular, what might go here?
https://github.com/romanz/electrs/pull/911/files#diff-d74a249fc203937952048161cdcb91480927c1994d7a3e6b1f41404b79962aa2R16

cryptoquick:
[...]

PR #911, opened 2023-07-23
Axum HTTP server feature

### LINKED ISSUE #820: Feature: HTTP RPC API
We're currently using Blockstream's Esplora with BDK, but it's quite heavy and out of date with current electrs.

I've considered making a crate called `electrs-http-proxy` that's just a reimplementation of the Esplora REST API that proxies to wrap electrs. Would this make sense to do as a separate crate?

An HTTP server will likely be opinionated and might complicate maintenance of electrs itself, and I'd be open to other implementations behind feature gating, but if I was doing it, I'd probably default to updating the interface to use axum, while referencing the hyper handler interface used by Blockstream.

**Additional context**
For context, a link to the Blockstream fork for Esplora:
https://github.com/Blockstream/electrs

Their REST request handler, using hyper:
https://github.com/Blockstream/electrs/blob/a808b51d0d9301fa82390b985c57551966001f9b/src/rest.rs#L615

I'm open to opinions, advice, especially of anything I might seem to be unaware of.

cryptoquick:
WIP

Closes #820 #904

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -3,6 +3,21 @@
 version = 3
 
 [[package]]
+name = "addr2line"
+version = "0.20.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "f4fa78e18c64fce05e902adecd7a5eed15a5e0a3439f7b0e169f0252214865e3"
+dependencies = [
+ "gimli",
+]
+
+[[package]]
+name = "adler"
+version = "1.0.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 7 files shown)


whfuyn:
Any update on this one?

cryptoquick:
We realized we already had access to the mempool API on the device, so it no longer makes sense for us to go this route.

knocte:
@cryptoquick what was left to do here? (was it still WIP?) My team might want to pick this up and finish it.

cryptoquick:
@knocte I was never able to get tokio to run the axum server alongside electrs for some reason. It also needs the remaining methods to be implemented.

PR #917, opened 2023-08-14
Allow existing mempool sync on SIGINT

romanz:
Otherwise, electrs cannot be gracefully stopped during the first mempool sync (currently [having >100k transactions](https://mempool.space/)).

### DIFF
--- a/src/mempool.rs
+++ b/src/mempool.rs
@@ -14,6 +14,7 @@ use serde::ser::{Serialize, SerializeSeq, Serializer};
 use crate::{
     daemon::Daemon,
     metrics::{Gauge, Metrics},
+    signals::ExitFlag,
     types::ScriptHash,
 };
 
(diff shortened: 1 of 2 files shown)


src/mempool.rs
```diff
@@ -116,21 +117,28 @@ impl Mempool {
...
-            })
-            .collect();
-        let added = entries.len();
-        for (txid, tx, entry) in entries {
-            self.add_entry(*txid, tx, entry);
+        let to_add: Vec<Txid> = to_add.into_iter().collect();
+        let mut added = 0;
+        for chunk in to_add.chunks(100) {
+            if exit_flag.poll().is_err() {
+                warn!("interrupted while syncing mempool");
```
Kixunil, line 124: Sorry for late comment but maybe this should be `info`? IIUC there's nothing wrong/dangerous about it.

romanz: Good catch, thanks!

PR #928, opened 2023-08-27
Add txid collision scanner

romanz:
```
$ time cargo run --release --example tx_collisions db/bitcoin
    Finished release [optimized] target(s) in 0.05s
     Running `target/release/examples/tx_collisions db/bitcoin`
prefix=4ee974b6585fb468 heights: 91722 91880
prefix=220fe15429d88599 heights: 91812 91842

real	2m37.073s
user	2m34.878s
sys	0m3.164s
```

PR #931, opened 2023-08-29
Concurrent block processing

RCasatta:
By changing the signature of the `for_blocks` function, avoiding the `FnMut` and returning data instead, we can process blocks in parallel.
This requires changing the code using the function because at the moment closures are using mutability in the context.

The status of the PR is very early, there are bugs sometimes with TODO,  reviewers should avoid in-depth review at the moment, just want to demonstrate feasibility and have early feedback on the concept.

On my machine, with a node with an empty mempool the time taken by
`./history.py 1FeexV6bAHb8ybZjqQMjJrcCrHGW9sb6uF --no-merkle-proofs`
is reduced from about 3.5s to 2.5s


src/p2p.rs
```diff
@@ -124,10 +128,11 @@ impl Connection {
                     );
                     Ok(block)
                 })?;
-                self.blocks_duration
-                    .observe_duration("process", || func(hash, block));
+                result.push((hash, block));
             }
-            Ok(())
+            let x: Vec<_> = result.into_par_iter().map(|(a, b)| func(a, b)).collect(); // TODO restore observe
```
RCasatta, line 133: should instead start processing as soon as the data arrives, instead of starting when all the data arrives

RCasatta:
[This comment](https://github.com/romanz/electrs/pull/931#discussion_r1309016534) still applies but could be addressed in a subsequent PR.
Other known TODOs should be addressed

src/tracker.rs
```diff
@@ -105,17 +105,18 @@ impl Tracker {
     ) -> Result<Option<(BlockHash, Transaction)>> {
         // Note: there are two blocks with coinbase transactions having same txid (see BIP-30)
         let blockhashes = self.index.filter_by_txid(txid);
-        let mut result = None;
-        daemon.for_blocks(blockhashes, |blockhash, block| {
-            if result.is_some() {
-                return; // keep first matching transaction
-            }
```
RCasatta, line 112: This optimization is currently removed by this MR, is BIP-30 the only case or there are others?

romanz: Indeed, currently there are only two cases of collisions - https://github.com/romanz/electrs/pull/928#issuecomment-1694755943

RCasatta: With this MR it's not granted blocks are processed in height order so it's not necessarily the first occurrence of the tx to be returned.
Since there are only 2 txids, would it be ok to hardcode those?

antonilol: currently there are these 2, but bip 30 allows txids to collide with existing ones when no new output points would overwrite existing ones in the utxo set (the reason why bip 30 was made, more on this in the bip document)

for optimization purposes i think you can assume no collisions are there (and hardcode the 2 known ones), if it still outputs _a_ transaction with that txid when there is in fact a collision. just no panic or ub

src/p2p.rs
```diff
@@ -124,10 +128,16 @@ impl Connection {
                     );
                     Ok(block)
                 })?;
-                self.blocks_duration
-                    .observe_duration("process", || func(hash, block));
+                result.push((hash, block));
```
romanz, line 131: I'm afraid it may result in OOM when there are many blocks to scan when subscribing to a "popular" address.

RCasatta: Yeah, that is a legitimate worry...
I tried to spawn as soon as the first block arrives, which would also be a performance improvement like I said in my first comment. But I have a compile issue I can't solve https://github.com/romanz/electrs/commit/9443888fd084f5562857d1ccdb3b5fddb8418faa . Rayon docs https://docs.rs/rayon/latest/rayon/fn.scope.html even has a specific section about it, but the only suggestion avoiding the move is not working.
I may try again or use a less elegant solution of processing by chunks of blocks (like 16 or so) with the par iter

RCasatta: @romanz should be fixed in the last commit (thanks for the suggestion @afilini)

RCasatta:
Is something needed on my end?
I see CI failures but I am not sure, any suggestion is welcome

src/db.rs
```diff
@@ -22,6 +23,16 @@ impl WriteBatch {
         self.spending_rows.sort_unstable();
         self.txid_rows.sort_unstable();
     }
+    pub(crate) fn merge(mut self, other: WriteBatch) -> Self {
+        self.header_rows.extend(other.header_rows.into_iter());
+        self.funding_rows.extend(other.funding_rows.into_iter());
+        self.spending_rows.extend(other.spending_rows.into_iter());
+        self.txid_rows.extend(other.txid_rows.into_iter());
+        if self.height < other.height {
+            self.tip_row = other.tip_row
```
romanz, line 32: Should we also update `self.height`?

RCasatta: I don't think so ->, https://github.com/romanz/electrs/pull/931#discussion_r1310484987

Which is not ideal at all, but I am not sure if there are other options, or if I should keep this and add a comment

romanz:
> I see CI failures but I am not sure, any suggestion is welcome

IIUC, it seems to be related to this PR... could you please take a look?

src/p2p.rs
```diff
@@ -91,43 +91,66 @@ impl Connection {
...
+                    })?;
+                    let func = &func;
+                    let blocks_duration = &self.blocks_duration;
+                    let send = send.clone();
+                    s.spawn(move |_| {
+                        let r = blocks_duration.observe_duration("process", || func(hash, block));
+                        let _ = send.send(r);
+                    });
+                }
+                let result: Result<Vec<_>, std::sync::mpsc::RecvError> =
```
romanz, line 140: Is it OK to block here (from `rayon` point of view)?
Asking, since when using [this patch](https://github.com/romanz/electrs/commit/36c38284f989d3786e1e9e1bc560431bb1d946ea), it seems that a task is being spawned, but seems to not run:
```
[2023-11-20T19:44:54.785Z DEBUG electrs::p2p] peer inventory: [Block(0x423cfe9f541e1425b02b93d8db56f8fecdcb0ce8f58d33102d205c73f6bb037e)]
[2023-11-20T19:44:54.786Z DEBUG electrs::p2p] got 1 new headers
[2023-11-20T19:44:54.786Z INFO  electrs::index] indexing 1 blocks: [111..111]
[2023-11-20T19:44:54.786Z DEBUG electrs::p2p] loading 1 blocks
[2023-11-20T19:44:54.786Z DEBUG electrs::p2p] spawning task for 423cfe9f541e1425b02b93d8db56f8fecdcb0ce8f58d33102d205c73f6bb037e
[2023-11-20T19:44:54.786Z DEBUG electrs::p2p] waiting for 1 blocks
[2023-11-20T19:44:54.786Z DEBUG electrs::p2p] processing 423cfe9f541e1425b02b93d8db56f8fecdcb0ce8f58d33102d205c73f6bb037e
[2023-11-20T19:44:54.786Z DEBUG electrs::index] writing 3 funding and 11 spending rows from 2 transactions, 1 blocks
[2023-11-20T19:44:54.843Z INFO  electrs::signals] notified via SIG10
[2023-11-20T19:44:54.866Z INFO  electrs::chain] chain updated: tip=423cfe9f541e1425b02b93d8db56f8fecdcb0ce8f58d33102d205c73f6bb037e, height=111
[2023-11-20T19:44:54.866Z DEBUG electrs::p2p] loading 1 blocks
[2023-11-20T19:44:54.866Z DEBUG electrs::p2p] spawning task for 423cfe9f541e1425b02b93d8db56f8fecdcb0ce8f58d33102d205c73f6bb037e
[2023-11-20T19:44:54.866Z DEBUG electrs::p2p] waiting for 1 blocks
<it seems to get stuck here>
[2023-11-20T19:45:48.880Z DEBUG electrs::server] 0: recv {"jsonrpc":"2.0","method":"blockchain.estimatefee","id":283,"params":[25]}
[2023-11-20T19:45:48.880Z DEBUG electrs::server] 0: recv {"jsonrpc":"2.0","method":"blockchain.estimatefee","id":284,"params":[10]}
[2023-11-20T19:45:48.880Z DEBUG electrs::server] 0: recv {"jsonrpc":"2.0","method":"blockchain.estimatefee","id":285,"params":[5]}
[2023-11-20T19:45:48.880Z DEBUG electrs::server] 0: recv {"jsonrpc":"2.0","method":"blockchain.estimatefee","id":286,"params":[2]}
[2023-11-20T19:45:48.889Z DEBUG electrs::server] 0: recv {"jsonrpc":"2.0","method":"mempool.get_fee_histogram","id":287}
```
Not sure, but it may be related to https://github.com/rayon-rs/rayon/issues/835... WDYT?

RCasatta: It seems a good catch! Did you have a chance to try the patch with `rayon::in_place_scope` instead of `rayon::scope` as suggested in the issue?

Another alternative maybe could be to use `Arc<Mutex<Vec<R>>>` instead of the channel (so the blocking happens inside the spawn)

RCasatta: Unfortunately, I can't emulate the deadlock locally (I can see the output of another debug print after recv), maybe it depends on the number of threads used by rayon.
I am going to push a temporary commit to use the CI for this

RCasatta: The CI is not happy with `in_place_scope` and neither with the Mutex version https://github.com/RCasatta/electrs/actions/runs/6941492681/job/18882464601

RCasatta: > I can't emulate the deadlock locally (I can see the output of another debug print after recv)

With locally I mean without the docker...

How can I see the logs with the docker command as it is done (`docker run -v $PWD/contrib/:/contrib -v $PWD/tests/:/tests --rm electrs:tests bash /tests/run.sh`) in the CI  ?

romanz: > Unfortunately, the CI is not happy with `in_place_scope` and neither with the Mutex version https://github.com/RCasatta/electrs/actions/runs/6941492681/job/18882464601

Please rebase over latest `master`, the above failure should be fixed by https://github.com/romanz/electrs/pull/939.

romanz: > How can I see the logs with the docker command as it is done (`docker run -v $PWD/contrib/:/contrib -v $PWD/tests/:/tests --rm electrs:tests bash /tests/run.sh`) in the CI ?

electrs logs are stored under `/data/electrs` (inside the container):
https://github.com/romanz/electrs/blob/a01d58c35d2f6aaea7a479eb508404e7ff3affdf/tests/run.sh#L47

RCasatta: Sorry for the dumb questions, but if the container dies after an error how to check inside the container?

romanz: Sorry, the logs are indeed unavailable when the container fails in CI...

I have added two more commits (8326249088f251e92fd6621710144ea57338ea17 & acaebd0494e8392d8feadf7b85fd15f4dd1d9111), so it should be possible to reproduce the issue locally by:
```
$ docker build -f Dockerfile.ci . --rm -t electrs:tests-bug
$ docker run -v $PWD/contrib/:/contrib -v /tmp/data:/data -v $PWD/tests/:/tests --rm electrs:tests-bug bash -x /tests/run.sh
```
The logs should be under `/tmp/data/` after the container exits.

RCasatta: Thanks, in the meantime I launched a couple of jobs on the CI.
The bad news is that the `in_place_scope` seems to deadlock the same https://github.com/RCasatta/electrs/actions/runs/6947160675
The good news is that the Mutex version isn't deadlocking https://github.com/RCasatta/electrs/actions/runs/6946996655
But before cleaning up the Mutex version I think it could be possible a mutex-free solution...

RCasatta: > But before cleaning up the Mutex version I think it could be possible a mutex-free solution...

I was considering to use a `Vec<OnceLock<R>>` for thread results, however `OnceLock` has MSRV 1.70
So, I changed the `Mutex<Vec<R>>` to `Vec<Mutex<Option<R>>>` so that there are no blocking call excluding the end of the rayon scope

romanz:
Thanks @RCasatta! Will review at the weekend :)

src/p2p.rs
```diff
@@ -90,43 +90,78 @@ impl Connection {
...
-                        .parsed_owned();
-                    ensure!(
-                        &header.block_hash_sha2()[..] == hash.as_byte_array(),
-                        "got unexpected block"
-                    );
-                    Ok(block)
-                })?;
-                self.blocks_duration
-                    .observe_duration("process", || func(hash, block));
+            let mut result = Vec::with_capacity(blockhashes_len);
```
romanz, line 113: If possible, can we use https://github.com/romanz/electrs/commit/5aba2a1928ba2d602dc2defd8d66fee2ef523d5b to simplify this method?

src/tracker.rs
```diff
@@ -105,17 +105,18 @@ impl Tracker {
...
+        let result = daemon
+            .for_blocks(blockhashes, |blockhash, block| {
+                let mut visitor = FindTransaction::new(txid);
+                match bsl::Block::visit(&block, &mut visitor) {
+                    Ok(_) | Err(VisitBreak) => (),
+                    Err(e) => panic!("core returned invalid block: {:?}", e),
+                }
+                visitor.tx_found().map(|tx| (blockhash, tx))
+            })?
+            .first()
```
romanz, line 117: Not sure, but shouldn't we handle the case when the first block doesn't contain a match (due to an index lookup false positive)? -> https://github.com/romanz/electrs/commit/2ff5f0092d62f5d05cdb4d314de0609978f7c882

romanz:
Can be relevant for https://github/romanz/bindex-rs:
```
$ time target/release/bindex-cli -a <(echo 1FeexV6bAHb8ybZjqQMjJrcCrHGW9sb6uF) -1 
[2025-07-11T12:59:33.475260Z INFO  bindex::address::cache] added 1 new addresses to watch
[2025-07-11T12:59:33.475291Z INFO  bindex::address] index DB: "db/bitcoin", node URL: http://localhost:8332
[2025-07-11T12:59:33.647440Z INFO  bindex::db] CF headers: 11542 files, 84.077002 MBs
[2025-07-11T12:59:33.648124Z INFO  bindex::db] CF script_hash: 639 files, 40857.575215 MBs
[2025-07-11T12:59:34.394605Z INFO  bindex::address] block=0000000000000000000052404cd3d31dfcfa2e89734fa24f3037812bce7fd7f5 height=905080 headers loaded
[2025-07-11T12:59:34.419088Z INFO  bindex::db] started auto compactions
[2025-07-11T12:59:34.423074Z INFO  bindex::address::cache] adding 643 history entries, 643 transactions, 558 headers to cache=""
[2025-07-11T12:59:34.555550Z INFO  bindex_cli] 1 address history: 643 entries, balance: 79957.26791529 BTC, UTXOs: 645 [2.559206ms]
╭──────────────────────────────────────────────────────────────────┬─────────────────────────┬────────┬────────┬─────────────┬────────────────╮
│ txid                                                             │ time                    │ height │ offset │ delta       │ balance        │
├──────────────────────────────────────────────────────────────────┼─────────────────────────┼────────┼────────┼─────────────┼────────────────┤
│ 709ad0c259b0cdba3abe44812394af3911d61f29eede92c7fc98d11a988a6928 │ 2025-07-11 04:04:03 UTC │ 905024 │    717 │ +0.00008063 │ 79957.26791529 │
│ 6f89bdcc1a906da722c82e417af5e8693b73f5da8d180c7f02d28c09598a6fcf │ 2025-07-10 02:35:28 UTC │ 904837 │   2517 │ +0.00000733 │ 79957.26783466 │
│ 7f46c4b6a79e53bf2a814d5f1fa0feb4920e6f26b5872f7ff7b958a9eb2a19bc │ 2025-07-09 00:13:20 UTC │ 904663 │    382 │ +0.00000548 │ 79957.26782733 │
│ dd5afafc3bb8282406cd885e592342175e33c25b2ef525f011b8f6cabe33648c │ 2025-07-05 23:23:36 UTC │ 904202 │    327 │ +0.00001000 │ 79957.26782185 │
│ 6a7967c70d2c8de5b6898df6b87b4eb785189b2cd6dad7f66cc29da343858e55 │ 2025-07-03 00:26:45 UTC │ 903744 │    371 │ +0.00000548 │ 79957.26781185 │
│ aace966bc4212e8910c3b96845776b86189ba3c4700ec9e635003b1a7926529c │ 2025-06-30 22:54:49 UTC │ 903448 │    641 │ +0.00000548 │ 79957.26780637 │
│ c45ad4f7ec533837c2bf68464ace6ee18a6383745495d892c99a146b20296f17 │ 2025-06-30 22:30:42 UTC │ 903447 │    579 │ +0.00000548 │ 79957.26780089 │
│ b8ea427d25f5d7d1b48645b2cf4c109eb79447ccd496c4d83062be1503642d32 │ 2025-06-24 02:54:08 UTC │ 902476 │    175 │ +0.00020000 │ 79957.26779541 │
│ 977dcd3580bb9d91d57c9f6024765eb8aafc1b7531f993fdd35873928013ac25 │ 2025-06-13 13:16:55 UTC │ 901082 │    604 │ +0.00001000 │ 79957.26759541 │
│ 2d6e71ed13b7201471b223d953d867806be1c0826f2c287c84be4547573bea99 │ 2025-05-27 14:50:02 UTC │ 898617 │   3018 │ +0.00002025 │ 79957.26758541 │
│                               ...                                │           ...           │  ...   │  ...   │     ...     │      ...       │
╰──────────────────────────────────────────────────────────────────┴─────────────────────────┴────────┴────────┴─────────────┴────────────────╯

real	0m1.198s
user	0m0.739s
sys	0m0.380s
```

Issue #936, opened 2023-10-01
Bug: Electrs shows misleading error message when blockchain.transaction.get does not find the transaction


cjackie:
**Describe the bug**
Electrs shows misleading and confusing error message when getting a non-existing transaction using "blockchain.transaction.get"

**Electrs version**
0.9.11.

**To Reproduce**
1. Configure and start electrs
2. Go to the `contrib` folder of the repo i.e. do "cd \<path to electrs\>/contrib"
3. Use ./get_tx.py to get a transaction with valid-looking transaction but does not exist in the Bitcoin network e.g. "dd8424a665e367d190b487a9bfb945420be78ff219a68ff81799c55c10fc6650"
5. See error in the response. The error is "message is No such mempool transaction. Use -txindex or provide a block hash to enable blockchain transaction queries. Use gettransaction for wallet transactions".

**Expected behavior**
The error indicating that the transaction is not found, instead showing the Bitcoin error message that suggests "txindex" is needed, or blockhash is provided. Setting txindex=1 in Bitcoin core is not required to use electrs (as advertised in README.md), and the blockhash should have been provided in https://github.com/romanz/electrs/blob/d0b69f0dc43f2d8ceefc8422f5231f15d064b699/src/electrum.rs#L370-L373

**Configuration**
Standard config

**Electrum client**
N/A

**Additional context**

cjackie:
Here is the fix:

change

```
      if verbose {
          let blockhash = self
              .tracker
              .lookup_transaction(&self.daemon, txid)?
              .map(|(blockhash, _tx)| blockhash);
          return self.daemon.get_transaction_info(&txid, blockhash);
      }
```

To

```
        if verbose {
            return match self.tracker.lookup_transaction(&self.daemon, txid)?.map(|(blockhash, _tx)| blockhash) {
                None => bail!("Transaction is not found."),
                Some(blockhash) => self.daemon.get_transaction_info(&txid, blockhash)
            }
        }
```

Then tighten the second parameter of daemon.get_transaction_info, making it non-optional

Eunovo:
I'm working on a PR for this. The error, `{"error":{"code":2,"message":"No such mempool transaction. Use -txindex or provide a block hash to enable blockchain transaction queries. Use gettransaction for wallet transactions."},"id":0,"jsonrpc":"2.0"}` comes from bitcoind. I want to parse the resulting error json and map the `code: 2` error to something like "Transaction not found"

antonilol:
[...]

PR #942, opened 2023-10-17
Improve error message

Eunovo:
This PR improves the error message returned for blockchain.transaction.get when the bitcoin daemon cannot find the transaction.
See #936

### DIFF
--- a/src/daemon.rs
+++ b/src/daemon.rs
@@ -325,3 +325,113 @@ where
         Err(err) => bail!("batch {} request failed: {}", name, err),
     }
 }
+
+macro_rules! define_error_codes {
+    ($($(#[doc = $doc:expr])* $name:ident = $value:expr,)*) => {
+        /// Bitcoin Core RPC error codes
+        /// * https://github.com/bitcoin/bitcoin/blob/master/src/rpc/protocol.h
+        #[derive(Debug, Clone, Copy)]
+        #[repr(i32)]
+        pub(crate) enum CoreRPCErrorCode {
+            $(
+                $(#[doc = $doc])*
+                $name = $value,
+            )*
+        }
+
+        impl CoreRPCErrorCode {
+            pub(crate) fn from_error_code(code: i32) -> Option<Self> {
+                Some(match code {
+                    $(
+                        $value => Self::$name,
+                    )*
+                    _ => return None,
+                })
+            }
+        }
+    }
+}
+
+define_error_codes! {
+    /// RpcInvalidRequest is mapped to HTTP_BAD_REQUEST (400).
+    /// It is not used for application-layer errors
+    RpcInvalidRequest = -32600,
+    /// RpcMethodNotFound is mapped to HTTP_NOT_FOUND (404).
+    /// It is not be used for application-layer errors.
+    RpcMethodNotFound = -32601,
+    RpcInvalidParams = -32602,
+    /// RpcInternalError is used for genuine errors in bitcoind
+    RpcInternalError = -32603,
+    RpcParseError = -32700,
+
+    /// std::exception thrown in command handling
+    RpcMiscError = -1,
+    /// Unexpected type was passed as parameter
+    RpcTypeError = -3,
+    /// Invalid address or key
+    RpcInvalidAddressOrKey = -5,
+    /// Ran out of memory during operation
+    RpcOutOfMemory = -7,
+    /// Invalid, missing or duplicate parameter
+    RpcInvalidParameter = -8,
+    /// Database error
+    RpcDatabaseError = -20,
+    /// Error parsing or validating structure in raw format
+    RpcDeserializationError = -22,
+    /// General error during transaction or block submission
+    RpcVerifyError = -25,
+    /// Transaction or block was rejected by network rules
+    RpcVerifyRejected = -26,
+    /// Transaction already in chain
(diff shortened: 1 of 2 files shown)


antonilol:
might be worth to define all error codes (in a new enum) to allow for easy future mapping of them to a different or more verbose message (like this pr)

Eunovo:
> might be worth to define all error codes (in a new enum) to allow for easy future mapping of them to a different or more verbose message (like this pr)

I have added all the rpc error codes defined in bitcoin's protocol.h file

antonilol:
The enum members don't need the `i32` fields, they will only every be set to one value
define them like this:
`    RpcInvalidRequest = -32600,`

and the enum definition and `parse_code` function can be macro generated, to not have 2 places where these codes are mapped to the enum member (but 1) to avoid inconsistencies/desyncs

Eunovo:
> The enum members don't need the `i32` fields, they will only every be set to one value define them like this: ` RpcInvalidRequest = -32600,`
> 
> and the enum definition and `parse_code` function can be macro generated, to not have 2 places where these codes are mapped to the enum member (but 1) to avoid inconsistencies/desyncs

@antonilol All done

src/daemon.rs
```diff
@@ -253,3 +253,54 @@ pub(crate) fn extract_bitcoind_error(err: &bitcoincore_rpc::Error) -> Option<&Rp
...
+    RpcWalletPassphraseIncorrect = -14,
+    RpcWalletWrongEncState = -15,
+    RpcWalletEncryptionFailed = -16,
+    RpcWalletAlreadyUnlocked = -17,
+    RpcWalletNotFound = -18,
+    RpcWalletNotSpecified = -19,
+    RpcWalletAlreadyLoaded = -35,
+    RpcWalletAlreadyExisits = -36,
+
+    RpcForbiddenBySafeMode = -2,
```
antonilol, line 305: can be removed as this was reserved in a version lower than the minimum supported bitcoind version of electrs

antonilol:
the documentation of the error codes can also be copied from the header file of bitcoind source
i meant something different with macro generated (`macro_rules! ...`), but maybe thats just me that i dont like introducing new dependencies. going from an enum variant to a number is as easy as the `as` keyword (`as i16` in this case), proc macros also add (sometimes significant) extra compile time
dont get me wrong, this works like i intended but adds a dependency

Eunovo:
> the documentation of the error codes can also be copied from the header file of bitcoind source

Are you referring to the comments in the protocol.h file?

> i meant something different with macro generated (`macro_rules! ...`), but maybe thats just me that i dont like introducing new dependencies. going from an enum variant to a number is as easy as the `as` keyword (`as i16` in this case), proc macros also add (sometimes significant) extra compile time dont get me wrong, this works like i intended but adds a dependency

I'm not sure I understand what you mean here. Can I generate a from_i16 implementation on the enum with just declarative macros (`macro_rules!`)? Even if I remove the num_traits dependency, it looks like I'll still have to create a proc_macro and that will require that I add the `syn` crate to pass the inputs to the macro.

antonilol:
> Are you referring to the comments in the protocol.h file?

yes

> Can I generate a from_i16 implementation on the enum with just declarative macros (`macro_rules!`)?

yes

with something along the lines of

```rust
macro_rules! define_error_codes {
    ($($name:ident = $value:expr,)*) => {
        /// some doc comments
        // some derives
        #[derive(Debug, Clone, Copy)]
        #[repr(i16)]
        pub enum CoreRPCErrorCode {
            $(
                $name = $value,
            )*
        }

        impl CoreRPCErrorCode {
            pub fn from_error_code(code: i16) -> Option<Self> {
                Some(match code {
                    $(
                        $value => Self::$name,
                    )*
                    _ => return None,
                })
            }
        }
    };
}

define_error_codes! {
    RpcInvalidRequest = -32600,
    RpcMethodNotFound = -32601,
    RpcInvalidParams = -32602,
    RpcInternalError = -32603,
    RpcParseError = -32700,

    ... etc
}

impl CoreRPCErrorCode {
    pub fn to_error_code(self) -> i16 {
        self as i16
    }
}
```

it can also be made more flexible but i dont think that is needed (at least yet), but you could also take the name and visibility of the enum as metavariables.

Eunovo:
@antonilol Thanks for your help with the `macro_rules!` definition. I've changed the implementation to use that and I also added doc comments. Please take a look and let me know if there's anything I need to change.

src/electrum.rs
```diff
@@ -653,7 +664,10 @@ impl Call {
                     .downcast_ref::<bitcoincore_rpc::Error>()
                     .and_then(extract_bitcoind_error)
                 {
-                    Some(e) => error_msg(&self.id, RpcError::DaemonError(e.clone())),
+                    Some(e) => match self.params.parse_rpc_error_code(e.code) {
+                        Some(err) => error_msg(&self.id, err),
```
Kixunil, line 668: IIUC this is sending the underlying error message to the caller. And while it's not recommended to connect untrusted devices to electrs some people may do that and it'd be better to just log the message and send back "internal server error" to avoid leaking information.

Eunovo: > IIUC this is sending the underlying error message to the caller.

It is not. I added `self.params.parse_rpc_error_code` to allow certain errors to be intercepted and replaced but the underlying error message is not sent to the caller

Issue #680, opened 2022-02-06
Turn off database log spam


rkfg:
I have `log_filters = "ERROR"` in the config and yet `electrs` produces enormous amount of useless database logs (in `db/bitcoin/LOG` file) like this:
```
2022/02/06-14:04:52.920247 7f96d12040 [db/db_impl/db_impl_compaction_flush.cc:1389] [txid] Manual flush finished, status: OK
2022/02/06-14:04:52.920252 7f96d12040 [db/db_impl/db_impl_compaction_flush.cc:1379] [funding] Manual flush start.
2022/02/06-14:04:52.920258 7f96d12040 [db/db_impl/db_impl_compaction_flush.cc:1389] [funding] Manual flush finished, status: OK
2022/02/06-14:04:52.920262 7f96d12040 [db/db_impl/db_impl_compaction_flush.cc:1379] [spending] Manual flush start.
2022/02/06-14:04:52.920268 7f96d12040 [db/db_impl/db_impl_compaction_flush.cc:1389] [spending] Manual flush finished, status: OK
2022/02/06-14:05:02.938064 7f96d12040 [db/db_impl/db_impl_compaction_flush.cc:1379] [config] Manual flush start.
2022/02/06-14:05:02.938215 7f96d12040 [db/db_impl/db_impl_compaction_flush.cc:1389] [config] Manual flush finished, status: OK
2022/02/06-14:05:02.938225 7f96d12040 [db/db_impl/db_impl_compaction_flush.cc:1379] [headers] Manual flush start.
2022/02/06-14:05:02.938231 7f96d12040 [db/db_impl/db_impl_compaction_flush.cc:1389] [headers] Manual flush finished, status: OK
2022/02/06-14:05:02.938237 7f96d12040 [db/db_impl/db_impl_compaction_flush.cc:1379] [txid] Manual flush start.
2022/02/06-14:05:02.938243 7f96d12040 [db/db_impl/db_impl_compaction_flush.cc:1389] [txid] Manual flush finished, status: OK
2022/02/06-14:05:02.938248 7f96d12040 [db/db_impl/db_impl_compaction_flush.cc:1379] [funding] Manual flush start.
2022/02/06-14:05:02.938253 7f96d12040 [db/db_impl/db_impl_compaction_flush.cc:1389] [funding] Manual flush finished, status: OK
2022/02/06-14:05:02.938258 7f96d12040 [db/db_impl/db_impl_compaction_flush.cc:1379] [spending] Manual flush start.
[...]

PR #959, opened 2023-11-24
Allow setting RocksDB log directory

romanz:
Following https://github.com/romanz/electrs/issues/680#issuecomment-1824841535.

### DIFF
--- a/internal/config_specification.toml
+++ b/internal/config_specification.toml
@@ -30,6 +30,12 @@ doc = "Directory to store index database (default: ./db/)"
 default = "\"./db\".into()"
 
 [[param]]
+name = "db_log_dir"
+type = "std::path::PathBuf"
+doc = "Directory to store index database internal log (default: same as specified by `db_dir`)"
+default = "\"\".into()"
+
+[[param]]
 name = "daemon_dir"
 type = "std::path::PathBuf"
 doc = "Data directory of Bitcoind (default: ~/.bitcoin/)"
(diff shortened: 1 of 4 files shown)


internal/config_specification.toml
```diff
@@ -29,6 +29,12 @@ type = "std::path::PathBuf"
 doc = "Directory to store index database (default: ./db/)"
 default = "\"./db\".into()"
 
+[[param]]
+name = "db_log_dir"
+type = "std::path::PathBuf"
+doc = "Directory to store index database internal log (default: same as specified by `db_dir`)"
+default = "\"\".into()"
```
Kixunil, line 36: I think it'd look better if we just made it optional ans only set it if it's `Some`. I'm surprised setting an empty even does the right thing.

Kixunil (commented):
Oh, lol, you already did this. Thanks! :D

internal/config_specification.toml
romanz: > I'm surprised setting an empty even does the right thing.

I agree, but it seems to work:
https://github.com/facebook/rocksdb/blob/f6fd4b9dbd15dba36f7e5ad23de407b5c26b1460/db/db_impl/db_impl_files.cc#L246

BTW, does `configure_me` support `Optional<PathBuf>`?

Kixunil: Yes, simply leave out `default = ""` and the type in the struct will be wrapped in `Option`. (For mandatory param you write `optional = false`.)

PR #963, opened 2023-12-01
Improve config documentation

### LINKED ISSUE #962: Config: Point to a specific configuration file
Is there any way to stop the service from wanting a configuration file at "/home/electrs/.electrs/config.toml."?

I am unable to point at a specific directory for reading the configuration file in my service file. Below is a copy of my service file.

`[Service]`
`ExecStart=/btc-server/electrs/bin/electrs --conf /etc/electrs/electrs.toml --log-filters INFO`
`Type=simple`

The service continuously fails to start because it wants to read a configuration file at /home/electrs/.electrs/config.toml. But there is no such file, it should just look at "/etc/electrs/electrs.toml".

Kixunil:
The `--skip-default-conf-files` flag is often desired but people may miss it because it's not in the doc, so this documents it.

Further, to avoid problems with people missing flags this adds a recommendation to run `electrs --help` so if we ever forget to update the doc people will learn about the new arguments anyway.

Fixes #962.

### DIFF
--- a/doc/config.md
+++ b/doc/config.md
@@ -25,6 +25,8 @@ $ bitcoind -server=1 -txindex=0 -prune=0
 ```
 ### Electrs configuration
 
+**Note:** this documentation may occasionally become stale. We recommend running `electrs --help` to get an up-to-date list of options.
+
 Electrs can be configured using command line, environment variables and configuration files (or their combination).
 It is highly recommended to use configuration files for any non-trivial setups since it's easier to manage.
 If you're setting password manually instead of cookie files, configuration file is the only way to set it due to security reasons.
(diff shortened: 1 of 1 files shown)


doc/config.md
```diff
@@ -25,6 +25,8 @@ $ bitcoind -server=1 -txindex=0 -prune=0
 ```
 ### Electrs configuration
 
+**Note:** this documenttion may occasionally become stale. We recommend running `electrs --help` to get an up-to-date list of options.
```
romanz, line 28:
```suggestion
**Note:** this documentation may occasionally become stale. We recommend running `electrs --help` to get an up-to-date list of options.
```

doc/config.md
```diff
@@ -38,6 +40,7 @@ can lead to serious problems! Currently the *only* permitted operation is *delet
 The Toml-formatted config files ([an example here](config_example.toml)) are (from lowest priority to highest): `/etc/electrs/config.toml`, `~/.electrs/config.toml`, `./electrs.toml`.
 
 The options in highest-priority config files override options set in lowest-priority config files.
+In case even an attempt to load these files is undesirable (common in case of protected systemd services) use the `--skip-default-conf-files` argument to prevent it.
```
romanz, line 43:
```suggestion
In case when an attempt to load these files is undesirable (common in case of protected systemd services), use the `--skip-default-conf-files` argument to prevent it.
```

romanz:
Thanks!

doc/config.md
romanz: Got it, thanks!

PR #964, opened 2023-12-03
Explain behavior of config file loading

Kixunil:
Previously it was not clear from the documentation that configuration files are optional or why non-not-exist errors are hard errors. This change adds the information about these important properties of config files to the documentation.


romanz:
Thanks!

PR #970, opened 2023-12-08
Async mempool scanning

### LINKED ISSUE #968: Initial mempool sync may take a lot of time
> All the above is strongly suggesting to me that the `rpc.sync()` call on [this line](https://github.com/romanz/electrs/blob/d84d7d672c456fceac55e8143d0775ed10684991/src/server.rs#L93-L94) is blocking for too long, probably because the `mempool.sync()` call on [this line](https://github.com/romanz/electrs/blob/d84d7d672c456fceac55e8143d0775ed10684991/src/tracker.rs#L77) is blocking. This results in JSON-RPC client calls not being handled because the electrs code relies on syncing being paused before handling incoming client requests.

@conduition Good catch, thanks!
It seems that the initial mempool sync takes quite a while... I'll open a new issue.

_Originally posted by @romanz in https://github.com/romanz/electrs/issues/961#issuecomment-1843453480_

conduition:
Fixes https://github.com/romanz/electrs/issues/968. Mempool scanning is now done asynchronously in a separate thread called the `mempool_sync` thread. 

## Method

My approach was to split the `Mempool::sync` method into two parts: (1) fetching the updated mempool transactions [slow, read-only] and (2) applying those transactions [fast, write].

- Mempool scanning is first triggered when the main thread sends a `HashSet<Txid>` to the `mempool_sync` thread. This set represents the current set of TXIDs in the mempool which we already know about. 
- The `mempool_sync` thread proceeds to scan the mempool from the remote node. While the scan is ongoing (could take a long time) the main thread can still process block updates, script notifications, and electrum client requests. 
- Once the mempool scan finishes, the `mempool_sync` thread blocks until it can send a `MempoolSyncUpdate` to the main thread. 
- When the main thread has a spare moment to receive this packaged update, it synchronously applies the update to the `Mempool`, via `Rpc::mempool_apply`. This adds new transactions, removes outdated ones, and updates the fee histogram. 

## Advantages

Scanning is asynchronous and read-only, while writing the update to the mempool's in-memory state is synchronous WRT the main-thread. This way, we don't introduce any mutexes because we don't need concurrent read/writes of `Mempool`. Electrum client requests are always processed with a consistent view of the mempool, even if a scan is ongoing and could finish at any time.

## Histogram Perf

While i was in there, [I improved the performance of applying mempool updates, by avoiding needless recomputation of the `FeeHistogram` from scratch every time](https://github.com/romanz/electrs/commit/9b5f925934de1986c803881cb866f3770a872175). Instead we dynamically update fee histogram bins according to which transactions have been added or removed. I added unit tests to demonstrate the logic is sound. If preferable, I can separate the `FeeHistogram` changes and package it into another PR (although it will probably conflict with this one).

### DIFF
--- a/src/mempool.rs
+++ b/src/mempool.rs
@@ -139,6 +139,10 @@ impl Mempool {
         &self.fees
     }
 
+    pub(crate) fn all_txids(&self) -> HashSet<Txid> {
+        HashSet::<Txid>::from_iter(self.entries.keys().copied())
+    }
+
     pub(crate) fn get(&self, txid: &Txid) -> Option<&Entry> {
         self.entries.get(txid)
     }
@@ -196,22 +200,6 @@ impl Mempool {
             self.vsize.set(&label, self.fees.vsize[bin_index] as f64);
             self.count.set(&label, self.fees.count[bin_index] as f64);
         }
-    }
-
-    pub fn sync(&mut self, daemon: &Daemon, exit_flag: &ExitFlag) {
-        let old_txids = HashSet::<Txid>::from_iter(self.entries.keys().copied());
-
-        let poll_result = MempoolSyncUpdate::poll(daemon, old_txids, exit_flag);
-
-        let sync_update = match poll_result {
-            Ok(sync_update) => sync_update,
-            Err(e) => {
-                warn!("mempool sync failed: {}", e);
-                return;
-            }
-        };
-
-        self.apply_sync_update(sync_update);
     }
 
     /// Add a transaction entry to the mempool and update the fee histogram.
--- a/src/mempool.rs
+++ b/src/mempool.rs
@@ -24,6 +24,22 @@ pub(crate) struct Entry {
     pub has_unconfirmed_inputs: bool,
 }
 
+impl Entry {
+    pub fn new(
+        txid: Txid,
+        tx: Transaction,
+        rpc_entry: bitcoincore_rpc::json::GetMempoolEntryResult,
+    ) -> Entry {
+        Entry {
+            txid,
+            tx,
+            vsize: rpc_entry.vsize,
+            fee: rpc_entry.fees.base,
+            has_unconfirmed_inputs: !rpc_entry.depends.is_empty(),
+        }
+    }
+}
+
 /// Mempool current state
 pub(crate) struct Mempool {
     entries: HashMap<Txid, Entry>,
@@ -38,8 +54,9 @@ pub(crate) struct Mempool {
 /// An update to [`Mempool`]'s internal state. This can be fetched
 /// asynchronously using [`MempoolSyncUpdate::poll`], and applied
 /// using [`Mempool::apply_sync_update`].
+#[derive(Default)]
 pub(crate) struct MempoolSyncUpdate {
-    new_entries: Vec<Entry>,
+    pub(crate) new_entries: Vec<Entry>,
     removed_entries: HashSet<Txid>,
 }
 
@@ -173,6 +190,11 @@ impl Mempool {
     pub fn apply_sync_update(&mut self, update: MempoolSyncUpdate) {
         let removed = update.removed_entries.len();
         let added = update.new_entries.len();
+
+        // Return early to avoid spurious logs.
+        if added == 0 && removed == 0 {
+            return;
+        }
 
         for txid_to_remove in update.removed_entries {
             self.remove_entry(txid_to_remove);
--- a/src/daemon.rs
+++ b/src/daemon.rs
@@ -220,6 +220,12 @@ impl Daemon {
             .context("failed to get mempool txids")
     }
 
+    pub(crate) fn get_mempool_entry(&self, txid: &Txid) -> Result<json::GetMempoolEntryResult> {
+        self.rpc
+            .get_mempool_entry(txid)
+            .context("failed to get mempool entry")
+    }
+
     pub(crate) fn get_mempool_entries(
         &self,
         txids: &[Txid],
--- a/src/electrum.rs
+++ b/src/electrum.rs
@@ -9,15 +9,17 @@ use rayon::prelude::*;
 use serde_derive::Deserialize;
 use serde_json::{self, json, Value};
 
-use std::collections::{hash_map::Entry, HashMap};
+use std::collections::{hash_map::Entry, HashMap, HashSet};
 use std::fmt;
 use std::iter::FromIterator;
 use std::str::FromStr;
+use std::sync::Arc;
 
 use crate::{
     cache::Cache,
     config::{Config, ELECTRS_VERSION},
     daemon::{self, extract_bitcoind_error, Daemon},
+    mempool::MempoolSyncUpdate,
     merkle::Proof,
     metrics::{self, Histogram, Metrics},
     signals::Signal,
@@ -123,7 +125,7 @@ pub struct Rpc {
     tracker: Tracker,
     cache: Cache,
     rpc_duration: Histogram,
-    daemon: Daemon,
+    daemon: Arc<Daemon>,
     signal: Signal,
     banner: String,
     port: u16,
@@ -147,7 +149,7 @@ impl Rpc {
             tracker,
             cache,
             rpc_duration,
-            daemon,
+            daemon: Arc::new(daemon),
             signal,
             banner: config.server_banner.clone(),
             port: config.electrum_rpc_addr.port(),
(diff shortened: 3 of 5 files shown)


romanz:
Many thanks!
Could you please open a separate PR for the first two commits?
(You can keep these commits also in this PR, to prevent merge conflicts)

src/mempool.rs
```diff
@@ -184,6 +179,28 @@ impl Mempool {
...
+        let limit = 1u128 << (FeeHistogram::BINS - bin_index - 1);
+        let label = format!("[{:20.0}, {:20.0})", limit / 2, limit);
+
+        if vsize_change >= 0 {
+            self.fees.insert(bin_index, vsize);
+        } else {
+            self.fees.remove(bin_index, vsize);
+        }
+
+        self.vsize.set(&label, self.fees.vsize[bin_index] as f64);
```
romanz, line 229: Can we skip label generation & Prometheus metrics' update during the initial sync?

src/mempool.rs
```diff
@@ -217,20 +234,38 @@ impl FeeHistogram {
...
+        }
+        if let Some(bin) = self.count.get_mut(bin_index) {
+            *bin += 1
+        }
+    }
+
+    fn remove(&mut self, bin_index: usize, vsize: u64) {
+        // skip transactions with too low fee rate (<1 sat/vB)
+        if let Some(bin) = self.vsize.get_mut(bin_index) {
+            *bin = bin.saturating_sub(vsize);
```
romanz, line 291: Maybe it's safer to use `bin.checked_sub(vsize).expect(...)`? 
(since we shouldn't have negative values here - unless there is a bug)

src/mempool.rs
```diff
@@ -144,13 +144,6 @@ impl Mempool {
...
             }
         }
-        self.fees = FeeHistogram::new(self.entries.values().map(|e| (e.fee, e.vsize)));
-        for i in 0..FeeHistogram::BINS {
-            let bin_index = FeeHistogram::BINS - i - 1; // from 63 to 0
-            let limit = 1u128 << i;
-            let label = format!("[{:20.0}, {:20.0})", limit / 2, limit);
-            self.vsize.set(&label, self.fees.vsize[bin_index] as f64);
-            self.count.set(&label, self.fees.count[bin_index] as f64);
         }
```
romanz, line 171: Please remove the trailing `}`.

src/mempool.rs
conduition: Sure, and probably log a `warn!` call if we somehow hit a `checked_sub` error. Fixed by amending dd8f994

src/mempool.rs
conduition: Woops, fixed by amending the commit.

src/mempool.rs
conduition: Absolutely. Actually now that I think about it, it doesn't make sense to update these labels for every single transaction. I'll push a fix which updates the metrics gauges only at the end of a mempool sync.

conduition:
Requested changes force pushed by amending the existing commits. @romanz [you can view the diffs here](https://github.com/romanz/electrs/compare/0bf77378ced2a91dc5b7fc45ba69f16a43dade40..0162c5fdf405e3c8d65af72844ee4bed05da11cb). If you feel comfortable with that, i'll submit https://github.com/romanz/electrs/pull/970/commits/9b5f925934de1986c803881cb866f3770a872175 and https://github.com/romanz/electrs/pull/970/commits/ed42680992994488961bbf5aa3028a7a3d4b5dcc as a new PR.

src/mempool.rs
conduition: Done in https://github.com/romanz/electrs/pull/970/commits/ed42680992994488961bbf5aa3028a7a3d4b5dcc

romanz:
> If you feel comfortable with that, i'll submit [9b5f925](https://github.com/romanz/electrs/commit/9b5f925934de1986c803881cb866f3770a872175) and [ed42680](https://github.com/romanz/electrs/commit/ed42680992994488961bbf5aa3028a7a3d4b5dcc) as a new PR.

Please do :)

conduition:
[PR open here](https://github.com/romanz/electrs/pull/971), this PR is now dependent on that one

romanz:
Please rebase over `master` (since #971 is merged).

conduition:
Rebase complete :+1:

romanz:
Seems that the CI failure is not related to https://github.com/romanz/electrs/pull/970/commits/06303d90d87100163f05b1bf2f0a72a7d280eb4f - I'll merge it via https://github.com/romanz/electrs/pull/973

romanz:
Please rebase over latest `master`.

romanz:
Could you please take a look at the failure?
You can reproduce it locally by running:
```
$ docker build -f Dockerfile.ci . --rm -t electrs:tests
$ docker run -v $PWD/contrib/:/contrib -v $PWD/tests/:/tests --rm electrs:tests bash -x /tests/run.sh
...
+ TXID=20521100abc8df819901ccc5f964059ffc02303c4b421dccbf43322928221a2b
+ echo ' * get_tx_status'
 * get_tx_status
++ electrum --regtest --wallet=data/electrum/wallet get_tx_status 20521100abc8df819901ccc5f964059ffc02303c4b421dccbf43322928221a2b
++ jq -c .
Transaction not in wallet.
+ test '' == '{"confirmations":0}'
...
```
IIUC, it seems that the broadcasted transaction is not being reported back to Electrum client.

romanz:
You can modify `tests/run.sh` to debug this issue, e.g.:
```diff
diff --git a/tests/run.sh b/tests/run.sh
index e19b4d2..b19d8cd 100755
--- a/tests/run.sh
+++ b/tests/run.sh
@@ -13,6 +13,7 @@ cleanup() {
   	kill $j
   	wait $j
   done
+  cat data/electrs/regtest-debug.log
 }
 trap cleanup SIGINT SIGTERM EXIT
 
@@ -68,6 +69,8 @@ NEW_ADDR=`$EL getunusedaddress`
 echo " * payto & broadcast"
 TXID=$($EL broadcast $($EL payto $NEW_ADDR 123 --fee 0.001 --password=''))
 
+sleep 20
+
 echo " * get_tx_status"
 test "`$EL get_tx_status $TXID | jq -c .`" == '{"confirmations":0}'
 ```

**Comment by conduition** [comment]
Rebased :white_check_mark:  

I'm now looking into the mempool missing TX error. My guess is that async mempool scanning results in a delayed reaction where `electrs` only receives the newly broadcasted TX back from bitcoind _after_ the current mempool scan iteration runs once or twice. 

Prior to this PR, the synchronous nature of mempool scanning meant that after broadcasting a transaction, the next iteration of the main thread's event loop triggered a mempool scan, and further client requests (like the subsequent TX lookup) weren't processed until the mempool scan was complete, so the request always got the most up-to-date view of the mempool possible after broadcasting.

I see two options to address this:
1. After a TX broadcast, immediately execute a synchronous mempool poll/update cycle in the main thread (blocking other client requests)
2. After a TX broadcast, add the new transaction into the `Mempool` state manually (using `Mempool::apply_sync_update`). 

I think option 2 is the more appropriate choice. I don't see any obvious associated negatives. The mempool sync thread should remove the broadcasted TX later if the entry we add manually is evicted for some reason.

Still waiting for the dockerfile to build for the moment. i'll return tomorrow with an update once I've verified my suspicions

**Comment by conduition** [comment]
@romanz sorry for the delay. My suspicions were correct, and I applied a fix which automatically appends the newly broadcasted TX to the mempool immediately upon broadcast.

**Comment by romanz** [comment]
@conduition Could you please review https://github.com/romanz/electrs/pull/979 which should also help when syncing large mempools?

**Comment by conduition** [comment]
@romanz Rebased now that https://github.com/romanz/electrs/pull/979 has been merged. I think this PR still has merit, because even if mempool scans are now much faster, there's still no need to block client requests on them.

**Comment by conduition** [comment]
Putting this briefly into Draft state; i found I was getting spurious logs after a07b70d.
```
[2023-12-24T18:21:13.575Z DEBUG electrs::mempool] 82650 mempool txs: 0 added, 0 removed
[2023-12-24T18:21:13.583Z DEBUG electrs::mempool] 82650 mempool txs: 0 added, 0 removed
[2023-12-24T18:21:13.621Z DEBUG electrs::mempool] 82650 mempool txs: 0 added, 0 removed
[2023-12-24T18:21:13.675Z DEBUG electrs::mempool] 82650 mempool txs: 0 added, 0 removed
[2023-12-24T18:21:13.682Z DEBUG electrs::p2p] loading 1 blocks
[2023-12-24T18:21:15.677Z DEBUG electrs::status] 1 transactions from 1 blocks
[2023-12-24T18:21:15.677Z DEBUG electrs::mempool] 82650 mempool txs: 0 added, 0 removed
[2023-12-24T18:21:15.684Z DEBUG electrs::mempool] 82650 mempool txs: 0 added, 0 removed
[2023-12-24T18:21:15.759Z DEBUG electrs::mempool] 82650 mempool txs: 0 added, 0 removed
[2023-12-24T18:21:15.764Z DEBUG electrs::mempool] 82650 mempool txs: 0 added, 0 removed
```

will fix

conduition:
OK should be fixed now

conduition:
Hey @romanz, just wanted to check in and see if I should take the time to rebase this or?

PR #971, opened 2023-12-11
avoid redundant recomputation of fee histogram bins

conduition:
Extracted from https://github.com/romanz/electrs/pull/970. This PR is a prereq of #970.

`FeeHistogram` was being recomputed from scratch on every call to `Mempool::sync`. This was unnecessary. Instead we can mutate the existing histogram bin state once on each new transaction added to or removed from the mempool.

Prometheus mempool metrics are only updated once per bin at the end of every sync.


romanz (commented):
Thanks!

PR #1042, opened 2024-05-24
fix deprecation and clippy warnings

antonilol:
`Transaction::txid()` was deprecated in the update to 0.32 (just renamed)


romanz (approved):

romanz:
Thanks!

PR #1045, opened 2024-05-25
WIP: try other kv stores to possibly replace RocksDB

antonilol:
depends on #1043 because fixed size datatypes made this pr easier
depends on #1125 because some database crates require a higher msrv than 1.63

a bit inspired by #765, only not trying to get a dozen DBs to work at once, and not using any dynamic dispatch

still experimental because it performed a bit worse than rocksdb in my initial try (indexing on the signet), but that could be due to other factors as well (both were running from the same bitcoind so one might have gotten prioritized maybe?)

there is a few TODOs still (like metrics from the db) and probably some configuration still needed to get more speed out of the hardware

the `Database` trait defines all things the rest of electrs can do with a database backend, so more databases can be added later, even a built in in-memory db maybe for automated testing or just quick regtest setup

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -19,7 +19,7 @@ jobs:
         os: [ubuntu-latest]
         build-args:
           [
-            --locked --no-default-features,
+            --locked --no-default-features --features rocksdb,
             --locked
           ]
         include:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -396,6 +396,7 @@ dependencies = [
  "parking_lot",
  "prometheus",
  "rayon",
+ "redb",
  "rust-rocksdb",
  "serde",
  "serde_derive",
@@ -860,6 +861,15 @@ checksum = "22e18b0f0062d30d4230b2e85ff77fdfe4326feb054b9783a3460d8435c8ab91"
 dependencies = [
  "crossbeam-deque",
  "crossbeam-utils",
+]
+
+[[package]]
+name = "redb"
+version = "2.6.3"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "8eca1e9d98d5a7e9002d0013e18d5a9b000aee942eb134883a82f06ebffb6c01"
+dependencies = [
+ "libc",
 ]
 
 [[package]]
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -183,6 +183,9 @@ name = "bitflags"
 version = "2.9.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1b8e56985ec62d17e9c1001dc89c88ecd7dc08e47eba5ec7c29c7b5eeecde967"
+dependencies = [
+ "serde",
+]
 
 [[package]]
 name = "block-buffer"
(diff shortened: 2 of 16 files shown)


antonilol:
redb needs at least rustc 1.66, and the Database trait in the pr uses generic associated types which were stabilized in 1.65 (redb also uses them)
1.66 is almost 1 and a half year old and [debian stable still has 1.63](https://packages.debian.org/source/stable/rustc)

yancyribbens:
why not sled?

antonilol:
stats so far:
db size on signet (block height 197272), rocksdb: 152M, redb: 447M (rocksdb uses compression, right?)
indexing time, rocksdb: 1 minute, redb: 6 minutes on ssd
on hdd rocksdb takes about the same time and appears to commit only 3 times, while redb commits way more often and was really slow

clearly I am doing something wrong here because redb's performance is comparable to that of rocksdb according to the project maintainer's [benchmarks](https://github.com/cberner/redb?tab=readme-ov-file#benchmarks)

could signet just be the wrong thing to test with? will redb perform better (compared to rocksdb) once rocksdb can't do (almost) everything in ram?

antonilol:
> why not sled?

I initially considered starting with redb or sled, both are in beta, but I chose redb because [its file format is stable](https://github.com/cberner/redb?tab=readme-ov-file#status), contrary to sled which has [warnings about possible manual migration in its readme](https://github.com/spacejam/sled?tab=readme-ov-file#known-issues-warnings)

I will try sled right now, hopefully I get it to work with similar performance compared to rocksdb

yancyribbens:
fwiw sled claims to be in alpha now

antonilol:
> fwiw sled claims to be in alpha now

they name beta in their readme, anyway this should only be an issue if there are bugs in there, and the manual migrations only become an issue when updating sled in the future, if they put a breaking change in a patch version (semver people wont be happy then) we just pin it to a previous (working) patch version

antonilol:
good news for sled: its minimum rust version is below that of electrs (redb would require a higher one) and I can work around not being able to use generic associated types, there is a workaround for that for old versions of rust
(if I get it working I will revert the change to rust-toolchain.toml, it's set higher now for redb)

Kixunil:
I hope we don't end up with sled. Manual migration would be very annoying for the users.

antonilol:
It shouldn't be annoying for electrs _users_, there is a bit of doc from sled on how to do migration, (import and export functions), in electrs we can make this an automatic migration. I think "manual" refers to users of sled (rust projects that use it as library, like electrs) need to manually add/change code for some updates.

Kixunil:
Still no migration is better than migration which takes time and resources.

antonilol:
Closing this because I could not get the performance boost I wanted. Some other databases did have a smaller footprint on the disk but I wanted something faster and possibly pure Rust.

yancyribbens:
On the crates page for reddb, they have a pretty interesting comparison of different kv stores.  Not sure if you looked at it yet: https://crates.io/crates/redb.  Indeed Rockdb showed the smallest memory foot print and lmdb the best performance (both C projects I think).  reddb which looks like a pure rust project has some pretty good benchmarks if you are to trust the measurements are actually representative.

antonilol:
Yes I saw that. Rocksdb is the clear winner on storage space and does 3x better too at batch writes. Redb does win at range reads according to the benchmark but I have not tested querying yet, only indexing so far.
You should look at a few specific benchmarks, building the index in electrs does batch writes and queries do range reads. More representative benchmarks would includes electrs of course like comparing indexing time and response time for a set of queries.
The [schema](https://github.com/romanz/electrs/blob/master/doc/schema.md) could also be a performance limiter, I tried to keep it the same in this PR, but changing it can improve performance. [Fulcrum](https://github.com/cculianu/Fulcrum) has an interesting db schema that uses random reading in files at specific offsets, which should be very fast on SSDs.

yancyribbens:
Somewhat related, but I think now that libbitcoinkernal is a thing, this project could drop that in, and then in that way, there would be no reason run bitcoin core.  Then the IBD could all be done as part of the same process instead of the current two step process of a) downloading the blk files with bitcoin core then b) compacting and indexing..

PR #1052, opened 2024-06-08
deprecate unused config option `timestamp`

### LINKED ISSUE #1044: Bug: unused config fields (silently ignored config flags)
**Describe the bug**

The config fields `daemon_dir` and `args` (in src/config.rs) and `timestamp` (in the generated config struct) are unused. I discovered this by running a very recent build of clippy (`cargo +nightly clippy`) on the code base instead of clippy using the configured version (1.63).
The `daemon_dir` and `timestamp` command line arguments (or in the config file) are silently ignored.

**Electrs version**
current master, 603830f1b921928089c558d90519b7ca0c0d4658

**To Reproduce**
2 ways:
- look at the clippy output (make sure to have the nightly toolchain installed with rustup): `cargo +nightly clippy`
- pass ignored flags to electrs binary (with or without `--timestamp`, timestamps are logged)

**Expected behavior**
assuming no future plans with the `daemon_dir` flag, that can be removed
if you ask me `timestamp` can just be removed, and the current behavior of showing timestamps can be kept as is
what is the current use of `args`? if I try to pass unknown flags I get an error, so I assume it is always empty then

rest of the issue template is not applicable to this as it is a code issue

antonilol:
closes #1044 

Feel free to edit any of the messages by directly committing (but review is still fine)

Is `std::process::exit(1);` desirable? This forces users to remove the flag with a nice message, in future releases they will get an error from the arg parser. When not exiting people might (and will) miss it, how many people look at the logs when booting a program? (thinking about one-click install full node distros for example)

### DIFF
--- a/src/config.rs
+++ b/src/config.rs
@@ -146,7 +146,6 @@ pub struct Config {
     pub disable_electrum_rpc: bool,
     pub server_banner: String,
     pub signet_magic: Magic,
-    pub args: Vec<String>,
 }
 
 pub struct SensitiveAuth(pub Auth);
@@ -195,7 +194,7 @@ impl Config {
     pub fn from_args() -> Config {
         use internal::ResultExt;
 
-        let (mut config, args) =
+        let (mut config, _args) =
             internal::Config::including_optional_config_files(default_config_files())
                 .unwrap_or_exit();
 
@@ -290,6 +289,25 @@ impl Config {
             unsupported => unsupported_network(unsupported),
         }
 
+        let mut deprecated_options_used = false;
+
+        if config.timestamp {
+            eprintln!(
+                "Error: `timestamp` is deprecated, timestamps on logs is (and was) always \
+                enabled, please remove this option."
+            );
+            deprecated_options_used = true;
+        }
+
+        if config.verbose > 0 {
+            eprintln!("Error: please use `log_filters` to set logging verbosity",);
+            deprecated_options_used = true;
+        }
+
+        if deprecated_options_used {
+            std::process::exit(1);
+        }
+
         let daemon_dir = &config.daemon_dir;
         let daemon_auth = SensitiveAuth(match (config.auth, config.cookie_file) {
             (None, None) => Auth::CookieFile(daemon_dir.join(".cookie")),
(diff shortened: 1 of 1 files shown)


src/config.rs
```diff
@@ -282,17 +280,44 @@ impl Config {
             ResolvAddr::resolve_or_exit,
         );
 
+        let changed_daemon_dir = config.daemon_dir != default_daemon_dir();
+
+        let mut default_cookie_path = config.daemon_dir;
```
Kixunil, line 285: Oh, wait, if daemon dir affects cookie maybe we should keep using it?

src/config.rs
```diff
@@ -282,17 +280,44 @@ impl Config {
...
+        if changed_daemon_dir {
+            eprintln!(
+                "Error: `daemon_dir` is not used anymore and deprecated, \
+                please remove this option."
+            );
+            if config.auth.is_none() && config.cookie_file.is_none() {
+                eprintln!(
+                    "Set the `cookie_file` option to \"{}\" because \
+                    `daemon_dir` will be removed in a future release.",
+                    default_cookie_path.display()
```
Kixunil, line 305: Very nice messages!

src/config.rs
```diff
@@ -282,17 +280,44 @@ impl Config {
...
+                );
+            }
+            std::process::exit(1);
```
Kixunil, line 308: I was going to suggest warning instead but since the messages are extra nice exit is better to ensure people see the nice ones. Also might make sense to exit after both messages were printed.

src/config.rs
antonilol: it only affects the default cookie location, and that is only used when no cookie and no username and password are set

src/config.rs
antonilol: > I was going to suggest warning instead but since the messages are extra nice exit is better to ensure people see the nice ones.

thought that as well

> Also might make sense to exit after both messages were printed.

done

src/config.rs
antonilol: thank you :)

src/config.rs
Kixunil: Yes, it could be somewhat useful for people who don't care to figure out the exact location of the cookie. But I could be convinced this is a too weird feature to support.

romanz: Let's keep `daemon_dir` for now, and only remove `timestamp` in the PR.

romanz (approved):
Thanks!

romanz:
@antonilol please squash the commits and update PR's title.

antonilol:
Done!

romanz:
Thanks!

PR #1056, opened 2024-06-19
offer alternative methods for customizing TX broadcast privacy

conduition:
Transactions which are first seen being broadcast from a user's node are more likely to be heuristically associated with the node runner personally. To break this heuristic, node runners might choose to broadcast a TX indirectly through alternative channels to disassociate that TX from the bitcoin node. Unfortunately bitcoind does not offer this option natively (yet).

This PR adds two new ways for users to broadcast transactions submitted to an `electrs` server:

1. [The `pushtx` crate](https://github.com/alfred-hodler/pushtx). This initiates a short-lived set of P2P connections which publishes a transaction to random peers resolved from DNS seeds. The most private way to do this is using TOR, hiding the IP address of the sender. 

2. A custom script option. This allows users to customize the behavior of electrs broadcasting. It should return a zero status if the broadcast was successful, and emit an error message with a non-zero status otherwise.

### DIFF
--- a/doc/config_example.toml
+++ b/doc/config_example.toml
@@ -33,7 +33,8 @@ log_filters = "INFO"
 # How to broadcast transactions.
 # - "rpc" : use bitcoind to broadcast the TX directly
 # - "pushtx-clear" : send the transaction to random bitcoin peers resolved from DNS seeds directly over clearnet
-#                    (bad for privacy, don't use this unless you know what you're doing)
+#                    (bad for privacy, don't use this unless you have protected the outbound connections via
+#                    other means, such as a VPN)
 # - "pushtx-tor" : send the transaction to random bitcoin peers resolved from DNS seeds over a TOR proxy
 # - "script" : pass the hex-encoded transaction to a script (specified by tx_broadcast_script)
 tx_broadcast_method = "rpc"
--- a/doc/config_example.toml
+++ b/doc/config_example.toml
@@ -29,3 +29,15 @@ electrum_rpc_addr = "127.0.0.1:50001"
 
 # How much information about internal workings should electrs print. Increase before reporting a bug.
 log_filters = "INFO"
+
+# How to broadcast transactions.
+# - "rpc" : use bitcoind to broadcast the TX directly
+# - "pushtx-clear" : send the transaction to random bitcoin peers resolved from DNS seeds directly over clearnet
+# - "pushtx-tor" : send the transaction to random bitcoin peers resolved from DNS seeds over a TOR proxy
+# - "script" : pass the hex-encoded transaction to a script (specified by tx_broadcast_script)
+tx_broadcast_method = "rpc"
+
+# A script which is invoked with a hex-encoded transaction as the sole positional argument. Should exit
+# with a zero status if the broadcast was successful.
+#
+# tx_broadcast_script = "/usr/local/bin/broadcast.sh"
--- a/doc/config_example.toml
+++ b/doc/config_example.toml
@@ -33,6 +33,7 @@ log_filters = "INFO"
 # How to broadcast transactions.
 # - "rpc" : use bitcoind to broadcast the TX directly
 # - "pushtx-clear" : send the transaction to random bitcoin peers resolved from DNS seeds directly over clearnet
+#                    (bad for privacy, don't use this unless you know what you're doing)
 # - "pushtx-tor" : send the transaction to random bitcoin peers resolved from DNS seeds over a TOR proxy
 # - "script" : pass the hex-encoded transaction to a script (specified by tx_broadcast_script)
 tx_broadcast_method = "rpc"
--- a/src/config.rs
+++ b/src/config.rs
@@ -401,28 +401,25 @@ impl Config {
             std::process::exit(0);
         }
 
-        if config.tx_broadcast_script.is_some()
-            && config.tx_broadcast_method != TxBroadcastMethodConfigOption::Script
-        {
-            eprintln!(
-                "Error: tx_broadcast_script must not be configured if tx_broadcast_method != \"script\""
-            );
-            std::process::exit(1);
-        }
-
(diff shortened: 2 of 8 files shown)


src/config.rs
```diff
@@ -352,6 +411,8 @@ impl Config {
             sync_once: config.sync_once,
             skip_block_download_wait: config.skip_block_download_wait,
             disable_electrum_rpc: config.disable_electrum_rpc,
+            tx_broadcast_method: config.tx_broadcast_method,
+            tx_broadcast_script: config.tx_broadcast_script,
```
Kixunil, line 415: It'd be better to have another internal `enum TxBroadcastMethod` that carries `Script(String)` variant and do the conversion where you do the checks.

Kixunil (commented):
Personally, I doubt that sending a tx to random peers over clearnet helps. If a node detects another node connecting to it, then just sending a single tx and then disconnecting it can just assume the transaction is being broadcasted by the connecting client. So in effect it likely *leaks* privacy. As such I would probably refrain from adding it.

You need Dandelion for this to work privately.

conduition:
> I doubt that sending a tx to random peers over clearnet helps

@Kixunil I strongly agree, and that's why the `pushtx-tor` option is preferable for most people. But it's possible that the user's networking setup already accounts for this, and so we shouldn't force them to use a TOR localhost proxy if they don't want to. For instance, `electrs` may be running on a machine with a VPN connection which the user relies on for network-level TX broadcasting privacy, while the `bitcoind` node runs on a separate machine without such protections. 

Perhaps I could add some extra warnings to the `pushtx-clear` option to make it clear this option will degrade privacy if used incorrectly?

src/config.rs
conduition: I gave it a try with this commit: d695a222e95ec775bd4976d82f04abab52c4639f

i think adding another type makes the namespacing a bit messier as now we have `pub` types `TxBroadcastMethod`, `TxBroadcastMethodConfigOption`, and `TxBroadcaster`. But it does help us avoid an `unwrap` call in `electrum.rs`, so either works for me.

conduition:
Added extra warnings for `pushtx-clear`: https://github.com/romanz/electrs/pull/1056/commits/36b7bc48b47b33ae6b68c17e91f9cf639cbc4a05

Kixunil:
> But it's possible that the user's networking setup already accounts for this, and so we shouldn't force them to use a TOR localhost proxy if they don't want to.

Good point but this should be explicitly documented, so good that you did. FTR this is also useful on Whonix I think.

src/config.rs
```diff
@@ -346,6 +401,30 @@ impl Config {
...
+                    );
+                    std::process::exit(1);
+                }
+                Some(script_path) => TxBroadcastMethod::Script(script_path),
+            },
+            TxBroadcastMethodConfigOption::BitcoinRPC => TxBroadcastMethod::BitcoinRPC,
+            TxBroadcastMethodConfigOption::PushtxClear => TxBroadcastMethod::PushtxClear,
+            TxBroadcastMethodConfigOption::PushtxTor => TxBroadcastMethod::PushtxTor,
+        };
+
```
Kixunil, line 427:
```suggestion
        match (config.tx_broadcast_method, config.tx_broadcast_script) {
            (TxBroadcastMethodConfigOption::Script, Some(script_path)) => TxBroadcastMethod::Script(script_path),
            (TxBroadcastMethodConfigOption::Script, None) => {
                eprintln!("Error: tx_broadcast_script must be configured if tx_broadcast_method == \"script\"");
                std::process::exit(1);
            },
            (TxBroadcastMethodConfigOption::BitcoinRPC, None) => TxBroadcastMethod::BitcoinRPC,
            (TxBroadcastMethodConfigOption::PushtxClear, None) => TxBroadcastMethod::PushtxClear,
            (TxBroadcastMethodConfigOption::PushtxTor, None) => TxBroadcastMethod::PushtxTor,
            (_, Some(_)) => {
                eprintln!("Error: tx_broadcast_script is only allowed when tx_broadcast_method is \"script\"");
                std::process::exit(1);
            },
        };

```

doc/config_example.toml
```diff
@@ -29,3 +29,16 @@ electrum_rpc_addr = "127.0.0.1:50001"
 
 # How much information about internal workings should electrs print. Increase before reporting a bug.
 log_filters = "INFO"
+
+# How to broadcast transactions.
+# - "rpc" : use bitcoind to broadcast the TX directly
+# - "pushtx-clear" : send the transaction to random bitcoin peers resolved from DNS seeds directly over clearnet
+#                    (bad for privacy, don't use this unless you know what you're doing)
```
Kixunil, line 36:
```suggestion
#                    (bad for privacy, don't use this unless you have protected the outbound connections via other means such as VPN)
```

src/config.rs
conduition: Nice. Added

conduition:
Thanks for the suggestions @Kixunil, added both :slightly_smiling_face:

src/electrum.rs
```diff
@@ -360,7 +376,7 @@ impl Rpc {
     fn transaction_broadcast(&self, (tx_hex,): &(String,)) -> Result<Value> {
         let tx_bytes = Vec::from_hex(tx_hex).context("non-hex transaction")?;
         let tx = deserialize(&tx_bytes).context("invalid transaction")?;
-        let txid = self.daemon.broadcast(&tx)?;
+        let txid = self.tx_broadcaster.broadcast(&tx)?;
```
antonilol, line 379: Why use `Arc<Daemon>` instead of just `Daemon`? A `&Daemon` can be passed to the `broadcast` method here.

conduition:
One minor annoyance is that the `pushtx` crate v0.4.0 relies on `bitcoin@0.31` whereas electrs uses `bitcoin@0.32`. Please let me know if that is a problem and I'll file a PR upstream to `pushtx` for a version bump.

PR #1060, opened 2024-07-12
Don't fail if bitcoind fee estimation is disabled

### LINKED ISSUE #1051: Bug: Electrs fails when bitcoind has blocksonly=1
**Describe the bug**
I have both bitcoind and electrs fully synced in my server. I wanted now to enable `blocksonly=1` on bitcoind to limit the resources and bandwidth used by the node.

After enabling that option, electrs fails to connect and gives the following error:

```
2024-06-05T14:45:54.501Z WARN  electrs::electrum] RPC blockchain.estimatefee failed: failed to estimate fee: JSON-RPC error: RPC error response: RpcError { code: -32603, message: "Fee estimation disabled", data: None }
```

**Electrs version**
v0.10.5

**To Reproduce**
1. Configure and start bitcoind
2. Configure and start electrs
3. Wait sync of both servers to finish
4. Enable blocksonly=1 in bitcoind config 
3. Connect with electrum client (I'm using 4.5.2)
5. The client will never connect succesfully and the server will start logging the error above

**Expected behavior**
I expected that electrs would work fine with `blocksonly=1` option since other electrum server (like Electrum Personal Server for example) seem to work fine with that option enabled

**Configuration**
<details>
<summary>electrs.toml</summary>

```
# The listening RPC address of bitcoind, port is usually 8332
# Connect to bitcoind container on Docker
daemon_rpc_addr = "bitcoin:8332"

# The listening P2P address of bitcoind, port is usually 8333
# Connect to bitcoind container on Docker
daemon_p2p_addr = "bitcoin:8333"

# Directory where the index should be stored. It should have at least 70GB of free space.
db_dir = "/electrs/data/"

# bitcoin means mainnet. Don't set to anything else unless you're a developer.
network = "bitcoin"

# The address on which electrs should listen. Warning: 0.0.0.0 is probably a bad idea!
# Tunneling is the recommended way to access electrs remotely.
electrum_rpc_addr = "0.0.0.0:50001"

# How much information about internal workings should electrs print. Increase before reporting a bug.
log_filters = "INFO"
```

</details>

<details>
<summary>bitcoind.conf</summary>

```
# [core]

# Only download and relay blocks - ignore unconfirmed transaction
blocksonly=1

# Set database cache size in MB; machines sync faster with a larger cache. Recommend setting to 4000 if RAM is available.
dbcache=250

# Maintain a full transaction index, used by the getrawtransaction rpc call.
txindex=1

# [network]

# Maintain at most N connections to peers.
maxconnections=64

# [rpc]

# Accept command line and JSON-RPC commands.
server=1

# Bind to given address to listen for JSON-RPC connections. This option is ignored unless -rpcallowip is also passed. Port is optional and overrides -rpcport. Use [host]:port notation for IPv6. This option can be specified multiple times. (default: 127.0.0.1 and ::1 i.e., localhost)
rpcbind=0.0.0.0

# Allow JSON-RPC connections from specified source. Valid for <ip> are a single IP (e.g. 1.2.3.4), a network/netmask (e.g. 1.2.3.4/255.255.255.0) or a network/CIDR (e.g. 1.2.3.4/24). This option can be specified multiple times.
rpcallowip=0.0.0.0/0

# Listen for JSON-RPC connections on this port
rpcport=8332

# Username and hashed password for JSON-RPC connections. The field <userpw> comes in the format: <USERNAME>:<SALT>$<HASH>. RPC clients connect using rpcuser=<USERNAME>/rpcpassword=<PASSWORD> arguments. You can generate this value at https://jlopp.github.io/bitcoin-core-rpc-auth-generator/. This option can be specified multiple times.
rpcauth=...
```

</details>

**System running electrs**
 - Deployment method: Docker
 - OS name and version (docker image): mu373/docker-electrs

**Electrum client**
Client name (if not upstream desktop Electrum) and version: 4.5.2

romanz:
Fixes #1051.

### DIFF
--- a/src/daemon.rs
+++ b/src/daemon.rs
@@ -148,11 +148,15 @@ impl Daemon {
     }
 
     pub(crate) fn estimate_fee(&self, nblocks: u16) -> Result<Option<Amount>> {
-        Ok(self
-            .rpc
-            .estimate_smart_fee(nblocks, None)
-            .context("failed to estimate fee")?
-            .fee_rate)
+        let res = self.rpc.estimate_smart_fee(nblocks, None);
+        if let Err(bitcoincore_rpc::Error::JsonRpc(jsonrpc::Error::Rpc(RpcError {
+            code: -32603,
+            ..
+        }))) = res
+        {
+            return Ok(None); // don't fail when fee estimation is disabled (e.g. with `-blocksonly=1`)
+        }
+        Ok(res.context("failed to estimate fee")?.fee_rate)
     }
 
     pub(crate) fn get_relay_fee(&self) -> Result<Amount> {


src/daemon.rs
```diff
@@ -148,11 +148,14 @@ impl Daemon {
     }
 
     pub(crate) fn estimate_fee(&self, nblocks: u16) -> Result<Option<Amount>> {
-        Ok(self
-            .rpc
-            .estimate_smart_fee(nblocks, None)
-            .context("failed to estimate fee")?
-            .fee_rate)
+        let res = self.rpc.estimate_smart_fee(nblocks, None);
+        if let Err(bitcoincore_rpc::Error::JsonRpc(jsonrpc::Error::Rpc(ref e))) = res {
```
antonilol, line 152: I think you can use pattern matching here, like
```rust
if let Err(bitcoincore_rpc::Error::JsonRpc(jsonrpc::Error::Rpc(RpcError { code: -32603, .. }))) = res {
    ...
} else {
    ...
}
```

romanz: Good catch, thanks!

PR #1075, opened 2024-08-25
[POC/WIP] Add optional Bip352 silentpayments index

jlest01:
This PR proposes a new `--silent-payments-index` option that creates an index of Bip352 silent payments.
Most of the code is based on @Sosthene00's fork, but I made the following changes:

. It uses https://github.com/bitcoin-core/secp256k1/pull/1519 and https://github.com/rust-bitcoin/rust-secp256k1/pull/721 instead of the `rust-silentpayments` crate.
. It adds a new `--silent-payments-index` configuration option. Without this option, the server does not index silent payments and works the same as it does today.
. The code does not use `Box[u8]` in `db.rs`.

To run this, something like the following command can be used: `cargo run -- --log-filters=INFO --db-dir <db_dir> --daemon-dir ~/.bitcoin/ --network signet --electrum-rpc-addr="127.0.0.1:60001" --silent-payments-index`

### DIFF
--- a/src/index.rs
+++ b/src/index.rs
@@ -1,8 +1,10 @@
 use anyhow::{Context, Result};
 use bitcoin::consensus::{deserialize, Decodable, Encodable};
+use bitcoin::secp256k1::PublicKey;
 use bitcoin::hashes::Hash;
-use bitcoin::{BlockHash, OutPoint, Txid};
+use bitcoin::{BlockHash, OutPoint, Txid, XOnlyPublicKey};
 use bitcoin_slices::{bsl, Visit, Visitor};
+use std::collections::HashMap;
 use std::ops::ControlFlow;
 
 use crate::{
@@ -87,6 +89,7 @@ pub struct Index {
     chain: Chain,
     stats: Stats,
     is_ready: bool,
+    is_sp_ready: bool,
     flush_needed: bool,
 }
 
@@ -117,6 +120,7 @@ impl Index {
             chain,
             stats,
             is_ready: false,
+            is_sp_ready: false,
             flush_needed: false,
         })
     }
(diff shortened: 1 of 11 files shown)


securitybrahh:
does the node supports silent payments despite this PR is not merged?

src/index.rs
```diff
@@ -293,3 +407,141 @@ fn index_single_block(
...
+                    let outpoint = OutPoint {
+                        txid,
+                        vout: i.try_into().expect("Unexpectedly high vout"),
+                    };
+                    if self
+                        .index
+                        .store
+                        .iter_spending(SpendingPrefixRow::scan_prefix(outpoint))
+                        .next()
+                        .is_none()
```
antonilol, line 447: If I understand correctly, this must make sure only transactions with (P2TR) unspent outputs are scanned.
Making an index dependent on data from another index seems odd to me, I have some questions about this:
- Does this handle collisions of `spending` entries in the database (correctly)?
- Wouldn't a user want historic data after restoring a wallet?
- Looking at the code, `spending` is made sure to be updated to the current chain tip and then `tweaks` is indexed (do I get this correctly?). This will include spent transactions (imagine 1 P2TR output TXs) as blocks get propagated, but not on the initial sync or when spent in the same block. Is this intended?

src/index.rs
```diff
@@ -293,3 +407,141 @@ fn index_single_block(
...
+        daemon,
+        index,
+        map: &mut map
+    };
+    bsl::Block::visit(&block, &mut index_block).expect("core returned invalid block");
+    for (hash, tweaks) in map {
+        let height = index.chain.get_block_height(&hash).expect("Unexpected non existing blockhash");
+        let mut value: Vec<u8> = u64::try_from(height).expect("Unexpected invalid usize").to_be_bytes().to_vec();
+        value.extend(tweaks.iter());
+        batch.tweak_rows.push(value);
```
antonilol, line 541: If I understand correctly, the key of the stored key-value pair is the block height. Is RocksDB the right tool for the job then? I think this can easily be achieved using two files that only need appending when building/updating the index: one for the tweaks, and one for the offsets and lengths in the tweaks file, for every block (for random access).

romanz:
See https://github.com/romanz/bindex-rs/pull/105

Issue #709, opened 2022-04-26
Build: could not find `iterator` in `signal_hook` and SIGUSR1 not found in scope


vostrnad:
**Which command failed?**
`cargo build --locked --release` (with or without `--no-default-features`)

**What was the error message?**

<details>
<summary>Error message</summary>

```
error[E0432]: unresolved import `signal_hook::iterator`
 --> src\signals.rs:4:18
  |
4 | use signal_hook::iterator::Signals;
  |                  ^^^^^^^^ could not find `iterator` in `signal_hook`

error[E0425]: cannot find value `SIGUSR1` in this scope
  --> src\signals.rs:59:13
   |
59 |             SIGUSR1, // allow external triggering (e.g. via bitcoind `blocknotify`)
   |             ^^^^^^^ not found in this scope

error: aborting due to 2 previous errors
```

</details>

**System**
OS name and version: Windows 10
rustc 1.48.0 (7eac88abb 2020-11-16)
cargo 1.48.0 (65cbdd2dc 2020-10-14)

**Compilation**
Linking: static
Cross compilation: no
Target architecture: amd64

romanz:
Does this issue reproduces on latest `master`?

vostrnad:
Yes, the error message appears exactly the same.

romanz:
I guess it's related to https://docs.rs/signal-hook/latest/signal_hook/iterator/index.html not being supported on Windows.

romanz:
Maybe we can try to use https://crates.io/crates/ctrlc (which seems to support Windows).

seriouscatsoserious:
What about using tokio signal? Seems that ctrlc can't natively handle SIGUSR1

seriouscatsoserious:
https://github.com/romanz/electrs/pull/1122

Fixed in this PR

seriouscatsoserious:
Initial PR #1122 left an extraneous file used for local testing purposes in repo, deleted subsequently in #1123

PR #1123, opened 2024-12-10
Implement ctrlc in place of signal hook on Windows

seriouscatsoserious:
Satisfies #709, utilizing ctrlc in place of signal hook on Windows only.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -39,7 +39,7 @@ serde_json = "1.0"
 tiny_http = { version = "0.12", optional = true }
 
 [target.'cfg(windows)'.dependencies]
-ctrlc = "3.2.5"
+ctrlc = "=3.4.2" 
 
 [target.'cfg(not(windows))'.dependencies]
 signal-hook = "0.3"
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -36,8 +36,13 @@ rayon = "1.9"
 serde = "1.0"
 serde_derive = "1.0, <=1.0.171"  # avoid precompiled binaries (https://github.com/serde-rs/serde/issues/2538)
 serde_json = "1.0"
-signal-hook = "0.3"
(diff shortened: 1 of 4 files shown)


romanz:
Many thanks for the contribution!

romanz (changes requested):
Would it please be possible to add a CI job to make sure the new code builds as expected on Windows?

seriouscatsoserious:
Should note that there is no Windows job with --features metrics_process flag because Prometheus doesn't work on Windows and I didn't want to introduce more complexity just for that

Cargo.toml
```diff
@@ -36,9 +36,14 @@ rayon = "1.9"
 serde = "1.0"
 serde_derive = "1.0, <=1.0.171"  # avoid precompiled binaries (https://github.com/serde-rs/serde/issues/2538)
 serde_json = "1.0"
-signal-hook = "0.3"
 tiny_http = { version = "0.12", optional = true }
 
+[target.'cfg(windows)'.dependencies]
+ctrlc = "3.2.5"
```
romanz, line 42: Just from curiosity, why not using 3.4.2 `ctrlc`?
IIUC it should support our MSRV: https://github.com/Detegr/rust-ctrlc/blob/3.4.2/Cargo.toml#L15

seriouscatsoserious:
Updated to 3.4.2

rust-toolchain.toml
```diff
@@ -1,3 +1,3 @@
 [toolchain]
-channel = "1.63.0"
+channel = "1.69.0"
```
romanz, line 2: I would prefer to keep the MSRV to 1.63, to allow building this project using Debian stable (https://packages.debian.org/stable/rust/rustc).

seriouscatsoserious: Got it, done!

romanz (approved):
Many thanks!

PR #1125, opened 2024-12-17
Test on Debian Trixie & update RocksDB to 9.3.1

romanz:
Since https://github.com/rust-rocksdb/rust-rocksdb/ didn't support [Debian Trixie librocksdb 9.3.1 package](https://packages.debian.org/trixie/librocksdb-dev), let's try https://github.com/zaidoon1/rust-rocksdb/commit/6db8312ea71407733ccae025a1352fc2e2a2bd7f.


Mike12299 (approved):

romanz:
@seriouscatsoserious could you please try to see if it also work on Windows?

romanz:
Will be merged after https://github.com/romanz/electrs/pull/1203

romanz:
https://lists.debian.org/debian-devel-announce/2025/07/msg00003.html
> We are planning to release trixie on August 9th.

daywalker90:
I'm using dynmaic linking so after upgrading to debian 13 and rust 1.89 i tried to build this PR and i got 
```
warning: clang-sys@1.8.1: could not execute `llvm-config` one or more times, if the LLVM_CONFIG_PATH environment variable is set to a full path to valid `llvm-config` executable it will be used to try to find an instance of `libclang` on your system: "couldn't execute `llvm-config --prefix` (path=llvm-config) (error: No such file or directory (os error 2))"
error: failed to run custom build command for `clang-sys v1.8.1`

Caused by:
  process didn't exit successfully: `/home/bitcoin/electrs/target/release/build/clang-sys-d05a70b9147a9fe9/build-script-build` (exit status: 101)
  --- stdout
  cargo:warning=could not execute `llvm-config` one or more times, if the LLVM_CONFIG_PATH environment variable is set to a full path to valid `llvm-config` executable it will be used to try to find an instance of `libclang` on your system: "couldn't execute `llvm-config --prefix` (path=llvm-config) (error: No such file or directory (os error 2))"

  --- stderr

  thread 'main' panicked at /home/bitcoin/.cargo/registry/src/index.crates.io-1949cf8c6b5b557f/clang-sys-1.8.1/build/dynamic.rs:225:45:
  called `Result::unwrap()` on an `Err` value: "couldn't find any valid shared libraries matching: ['libclang.so', 'libclang-*.so'], set the `LIBCLANG_PATH` environment variable to a path where one of these files can be found (invalid: [])"
  note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
warning: build failed, waiting for other jobs to finish...
```

I thought maybe the upgrade uninstalled clang or there are new dependencies but 
`sudo apt install clang cmake build-essential`

showed it was still installed.

so i did what everyone does nowadays and asked an llm which told me to
`sudo apt install llvm-dev libclang-dev`
and the build worked after.

It's running and working now.

> DB format will be upgraded, so switching back to older electrs versions will require re-indexing.

I didn't see that in the logs

romanz:
> `sudo apt install llvm-dev libclang-dev` and the build worked after.

Thanks - will update the docs.



> > DB format will be upgraded, so switching back to older electrs versions will require re-indexing.

> I didn't see that in the logs, is that normal?

If you switch to the new RocksDB version, it should work (the old DB format can be read by the new code) without reindexing.

The other way (reading the new DB format using old RocksDB) won't work.

romanz:
CC: @Kixunil @openoms @rootzoll @mynodebtc @craigraw @Stadicus @lukechilds @nmfretz @gstart9

romanz:
Will be released as `0.11.0`

nmfretz:
Thanks for the ping @romanz

prusnak:
I confirm this builds in Nixpkgs with latest rocksdb there (10.4.2).

Geremia:
I get #623 when building this.

PR #1128, opened 2024-12-20
Fixup MSRV setting via `Cargo.toml`

romanz:
Also, fix a 'unused_imports' lint.


romanz:
The CI will use stable Rust for Build jobs.
Integration job is using MSRV (1.63.0) from Debian stable.

PR #1129, opened 2024-12-22
Separate reading & writing into scoped threads

romanz:
If the index is on a SSD, sync perfomance should improve.

When the blocks' data is stored on a HDD, it takes ~3h to finish the initial sync (without full compactions, which take an additional ~1h):

Most of the time is spent waiting for p2p messages (i.e. bitcoind reading blocks from the spinning disk):

Writing the index is done in parallel with the reading:


romanz:
Without this PR, the initial sync takes ~4h50m (excluding ~1h of full compaction).

Issue #282, opened 2020-07-25
Any thoughts on making an "official" docker hub image?


truesilver92:

romanz:
Good question - as of today I prefer to have source releases, since I don't want to require more trust when downloading a binary/Docker release.
You can use the included [Dockerfile](https://github.com/romanz/electrs/blob/master/Dockerfile) to build your own image if needed.

Kixunil:
Building `electrs` with `librocksdb` linked dynamically is deterministic, so we could both sign such binaries and anyone else interested could join. I already sign my deb packages but I understand most people will not want to manually unpack them (unless it becomes the preferred method of deployment), so I'm willing to sign binaries themselves if there's a demand.

thomaseizinger:
> Good question - as of today I prefer to have source releases, since I don't want to require more trust when downloading a binary/Docker release. You can use the included [Dockerfile](https://github.com/romanz/electrs/blob/master/Dockerfile?rgh-link-date=2020-07-25T09%3A45%3A42Z) to build your own image if needed.

Would you be interested in a GitHub workflow that builds and attaches binaries to every release? I would like to deploy electrs to my raspberry and was thinking of creating a `-bin` AUR package to avoid compiling on it. It would be significantly easier if I could just download a binary from the release.

I understand that this requires trust in GitHub's CI service but maybe we can use it as a starting point? Tooling to verify that the build is in fact what we expect and as a result uploading a signature can be built on top of this.

Kixunil:
@thomaseizinger if can get some people to reproduce and sign the releases then I'm in favor.

thomaseizinger:
> @thomaseizinger if can get some people to reproduce and sign the releases then I'm in favor.

I am not sure I completely follow. Uploading a release-asset requires write access to the repository so I think it would have to be done by maintainers? Or are you thinking of a workflow where non-maintainers reproduce and sign and then request the signature to be uploaded to the release?

Kixunil:
We could have some dedicated place to upload signatures for verified developers (to avoid spam).

SRv6d:
In the spirit of  "[SO WHY THE FUCK IS THERE CODE? make an EXE file and give it to me. STUPID FUCKING SMELLY NERDS](https://github.com/sherlock-project/sherlock/issues/2011)", can this be reevaluated?
I would argue that having to compile code before you can use it is quite the hurdle for the average user. Can we make it easier to run your own node by offering compiled binaries?
[...]

PR #1187, opened 2025-05-02
update

SwissBitcoinPay:


romanz (commented):
Are the generated artifacts deterministic?
Please see https://github.com/romanz/electrs/issues/282#issuecomment-663835475 for context.

PR #1191, opened 2025-05-16
Fix broken Docker installation link in upgrading guide

GarmashAlex:
Replaced the outdated reference to docker.md with the correct link to the Docker installation section in install.md within the upgrading guide. This ensures users are directed to the proper instructions for Docker-based installation and configuration.


romanz (approved):
Thanks!

PR #1217, opened 2025-08-24
Update `serde` to the latest version after the precompiled library got removed

antonilol:
I set it to 1.0.184 in Cargo.toml for the same reason it got done [here](https://github.com/RustCrypto/RSA/pull/360), Cargo will pull in the latest version actually (the now latest version 1.0.219 is semver-compatible with 1.0.184).

Serde's changelog mentions mostly bug fixes and binary size improvements, I only noticed a slight improvement in binary size (after this PR 99.97% of the size of before).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -1003,18 +1003,18 @@ dependencies = [
 
 [[package]]
 name = "serde"
-version = "1.0.171"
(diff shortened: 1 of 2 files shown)


romanz (approved):
Thanks!

PR #1220, opened 2025-08-25
cache cf handles in DBStore::write

antonilol:
Getting the column family handle makes rust-rocksdb do a `BTreeMap` lookup and electrs then checks if it exists with `.expect(...)`.
Doing this outside the for loops should improve performance, compared to doing it for every row that is inserted.

I will do some test runs (with prometheus metrics) tomorrow, would like to see what performance gains this gives exactly.

### DIFF
--- a/src/db.rs
+++ b/src/db.rs
@@ -284,19 +284,23 @@ impl DBStore {
 
     pub(crate) fn write(&self, batch: &WriteBatch) {
         let mut db_batch = rocksdb::WriteBatch::default();
+        let funding_cf = self.funding_cf();
         for key in &batch.funding_rows {
-            db_batch.put_cf(self.funding_cf(), key, b"");
+            db_batch.put_cf(funding_cf, key, b"");
         }
+        let spending_cf = self.spending_cf();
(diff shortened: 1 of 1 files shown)


antonilol:
I compiled this and the commit before this (called the binaries `before` and `after`), ran then 4 times each (alternating `before` and `after`) on the signet, storing the database in memory. I noticed an almost 2% speedup. The rust compiler probably does not optimize this because the `BTreeMap` is used in `rust-rocksdb`, and no inlining is done by default across crate boundaries.

The measurements for anyone interested:
```
run  binary  blocks  index_update_duration step="write" (s) 
1    before  267393  179.196
2    after   267393  175.306
3    before  267394  178.330
4    after   267397  175.907
5    before  267398  180.414
6    after   267398  175.544
7    before  267399  177.981
8    after   267400  179.004    outlier

avg time (s) compared to "before"
before: 178.98025   100.00%
 after: 176.44025    98.58%
        175.585667   98.10%     without outlier
```

romanz (approved):
Many thanks!

PR #1227, opened 2025-09-06
fix: outdated comment jsonrpc_timeout is for RPC calls, not wait_for_new_block

Snezhkko:
The comment incorrectly referenced a non-existent function. The timeout
actually controls RPC call duration, while new block waiting is handled
via new_block_rx channel in server.rs.


romanz (approved):
Thanks!

PR #1235, opened 2025-10-16
chore: fix some minor issues in the comments

vastonus:
Remove extra space in comment.


romanz (changes requested):

src/config.rs
```diff
@@ -48,7 +48,6 @@ impl fmt::Display for AddressError {
 }
 
 /// Newtype for an address that is parsed as `String`
-///
```
antonilol: Removing this will stick the lines together like "... as `String` The main point ...", that seems unintentional.

PR #1251, opened 2025-12-23
multi: expose configurable magic on non-signet

torkelrogstad:
There's no point in limiting this option to signet only. Allowing this flag to be set outside of signet networks lets electrs be used for custom testing networks that aren't signets.


romanz:
~Rebased and~ added a small simplification.

PR #1270, opened 2026-02-21
Use CDB for historic records

oren-z0:
RocksDB is a great key-value storage for mutable data, but most data in Bitcoin is immutable.
[CDB64](https://docs.rs/cdb64/latest/cdb64) is a very fast database for immutable data - using reference tables like this:

image

Originally CDB uses 32bits pointers, but cdb64 uses 64bits, source: https://www.unixuser.org/~euske/doc/cdbinternals/index.html.

Building the reference tables is a slow process, so the CDB can't replace RocksDB for fresh data. For example, we can store all the data up to block 900,000 in CDBs, and on query, search for it both in the CDB and the much-smaller RocksDB.

I haven't run benchmarks yet.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -228,6 +228,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "cdb64"
+version = "0.2.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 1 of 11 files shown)


romanz:
Closing in favor of https://github.com/romanz/bindex-rs/pull/99.

PR #1252, opened 2025-12-25
Switch to `bindex`

romanz:
Requires https://github.com/bitcoin/bitcoin/pull/33657 (to be released in Bitcoin Core 31).

For more information, see [slides](https://docs.google.com/presentation/d/1Zez-6DApKRu59kke4i_g9jwxQlaFKKRpOPdYFYsFXfA/).

PR #1288, opened 2026-05-20
Drop archived `bitcoincore-rpc` dependency

### LINKED ISSUE #1254: Drop `bitcoincore-rpc` dependency
Following https://github.com/rust-bitcoin/rust-bitcoincore-rpc/commit/92cad27319af78a77522090c4e1033ce7e738214.

PeterXMR:
Closes #1254.

The upstream `rust-bitcoincore-rpc` crate was [archived on 2025-11-25](https://github.com/rust-bitcoin/rust-bitcoincore-rpc/commit/92cad27319af78a77522090c4e1033ce7e738214). This PR removes it from electrs in favor of a small in-tree `rpc` module that calls bitcoind directly via the still-maintained `jsonrpc` crate (which electrs was already pulling in transitively and using directly in `batch_request`).

## Approach

I noticed `bitcoincore-rpc` was mostly a thin typed wrapper over `jsonrpc::Client`, and electrs uses only ~10 of its methods — so the replacement is narrow:

- **`Auth`** — local enum mirroring the three variants `config.rs` constructs.
- **`Client`** — wraps `jsonrpc::Client` and exposes the eight typed methods `Daemon` calls (`get_network_info`, `get_blockchain_info`, `estimate_smart_fee`, `send_raw_transaction`, `get_raw_transaction`, `get_block_info`, `get_mempool_info`, `get_raw_mempool`) plus the generic `call()` and `get_jsonrpc_client()` used for batch requests.
- **Response types** — `#[derive(Deserialize)]` structs containing **only the fields electrs actually reads**, matching the JSON-field renames bitcoind uses (`networkactive`, `initialblockdownload`, `relayfee`, `feerate`).
- **`Error`** — two-variant enum (`JsonRpc | Decode`) mirroring the nested shape callers already pattern-match against, so the changes in `daemon.rs` and `electrum.rs` are nearly mechanical.

## Diff

```
Cargo.lock      |  38 +-------- (bitcoincore-rpc + -rpc-json removed; no new deps)
Cargo.toml      |   2 +-        (drop bitcoincore-rpc, promote jsonrpc to direct)
src/config.rs   |   2 +-        (import swap)
src/daemon.rs   |  19 +++-----
src/electrum.rs |   2 +-        (downcast_ref target)
src/lib.rs      |   1 +         (mod rpc;)
src/rpc.rs      | 201 +++++++++ (new)
```

## Testing

- `cargo build --all-targets` — clean
- `cargo clippy --all-targets -- -D warnings` — zero warnings
- `cargo test --lib` — 21/21 pass (including `config::tests::test_auth_debug`, which exercises the local `Auth` enum's `Debug` formatting)
- **Integration smoke test** against Bitcoin Core 31 on regtest: drove every replaced RPC method end-to-end via the Electrum protocol:
  - `blockchain.relayfee` → `get_network_info::relay_fee` (BTC↔`Amount` serde)
  - `blockchain.estimatefee` → `estimate_smart_fee` (correctly maps missing `feerate` to `Option<Amount>::None` on regtest)
  - `blockchain.transaction.get` (hex) → `get_raw_transaction` (hex + consensus decode)
  - `blockchain.transaction.get` (verbose) → generic `Client::call`
  - `mempool.get_fee_histogram` → mempool sync exercising `get_mempool_info` + `get_raw_mempool` + batch `getmempoolentry` (incl. `fees.base`, `vsize`, `depends`) + batch `getrawtransaction`
  - `blockchain.transaction.broadcast` (fresh signed tx) → `send_raw_transaction`
  - error path: a deliberately-rejected rebroadcast returned through electrs as `{"code": 2, ...}`, confirming `extract_bitcoind_error` and the `downcast_ref::<crate::rpc::Error>()` in `electrum.rs` work as before.
- Implicitly exercised at startup: `get_blockchain_info` (`rpc_poll`), `get_block_info` (indexing 111 blocks), `get_mempool_info`.

## Notes

- I left `Auth::None` in place (marked `#[allow(dead_code)]`) for API completeness — `daemon.rs` matches it but no current code path constructs it. Easy to remove if you'd prefer.
- I deliberately did **not** add an `EstimateMode` enum (the old `bitcoincore_rpc_json::EstimateMode`) since electrs always passes `None` for it. The new `estimate_smart_fee` takes `Option<&str>` instead, which avoids dead variants. Happy to add the enum back if you'd rather keep the API surface.
- I'm aware of #1252 (`bindex`) — if that's expected to land soon and will rewrite most of `daemon.rs` anyway, this PR could be deferred or closed. It's a tidy interim cleanup either way.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -149,30 +149,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "bitcoincore-rpc"
-version = "0.19.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "aedd23ae0fd321affb4bbbc36126c6f49a32818dc6b979395d24da8c9d4e80ee"
-dependencies = [
- "bitcoincore-rpc-json",
- "jsonrpc",
- "log",
- "serde",
- "serde_json",
-]
-
-[[package]]
-name = "bitcoincore-rpc-json"
-version = "0.19.0"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d8909583c5fab98508e80ef73e5592a651c954993dc6b7739963257d19f0e71a"
-dependencies = [
- "bitcoin",
- "serde",
- "serde_json",
-]
-
-[[package]]
 name = "bitflags"
 version = "2.9.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
@@ -412,7 +388,6 @@ dependencies = [
  "bitcoin",
  "bitcoin-test-data",
  "bitcoin_slices",
- "bitcoincore-rpc",
  "configure_me",
  "configure_me_codegen",
  "crossbeam-channel",
@@ -420,6 +395,7 @@ dependencies = [
  "dirs-next",
  "env_logger",
  "hex_lit",
+ "jsonrpc",
  "log",
  "parking_lot",
  "prometheus",
@@ -595,7 +571,6 @@ source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "3662a38d341d77efecb73caf01420cfa5aa63c0253fd7bc05289ef9f6616e1bf"
 dependencies = [
  "base64",
- "minreq",
  "serde",
  "serde_json",
 ]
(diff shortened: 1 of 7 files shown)


romanz:
Done in #1252.

PR #1274, opened 2026-02-23
Option to use ZMQ+Rest to get blocks (as an alternative to p2p, the default) to make electrs work with nodes that connect through proxies

TheQuantumPhysicist:
Nodes that connect through proxies, with the `-proxy` command line argument, do not bind p2p, meaning, there is no 8333 port binding on loopback. This means that, by definition, electrs will not work on them, since it depends primarily on port p2p receive blocks.

This PR fixes this issue. It uses a combination of ZMQ, for new block notifications, and rest calls to the node to get blocks. The performance is comparable to p2p.

This was done by creating an abstraction layer, a trait, `BlockSource`, that abstracts both p2p and zmq/rest communication with the node.

To use this branch, make sure to:

1. Run your node with:`rest=1` and `zmqpubhashblock=tcp://127.0.0.1:28332` (config or command line args)
2. Run this branch of electrs with `--daemon-zmq-addr 'tcp://127.0.0.1:28332'`

I've been testing this for a while and it seems to be working. Would appreciate more testing.

Fixes #901

Issue #399, opened 2021-05-17
p2p-based sync fails to receive blocks


romanz:
Here's the output after updating to the latest commit:
```
Config { network: Bitcoin, db_path: "/home/electrs/bitcoin", daemon_dir: "/home/bitcoin/.bitcoin", daemon_cookie_file: "/home/bitcoin/.bitcoin/.cookie", daemon_rpc_addr: V4(127.0.0.1:8332), daemon_p2p_addr: V4(127.0.0.1:8333), electrum_rpc_addr: V4(0.0.0.0:50001), monitoring_addr: V4(127.0.0.1:4224), wait_duration: 30s, index_batch_size: 12, ignore_mempool: false, server_banner: "Welcome to electrs 0.9.0 (Electrum Rust Server)!", args: [] }
Error: initial sync failed
Caused by:
    0: failed to get block 00000000000006361896db3f164a5df9e83b2aa4f636e3ffac584d5b8c9c6c0c
    1: p2p failed to recv
    2: I/O error: unexpected end of file
```
If I let it run, it appears to keep failing on the same block.

I deleted the db dir (the service had been running, and failing, since I posted my issue), and then the block giving the error changed to this one:
`0: failed to get block 00000000839a8e6886ab5951d76f411475428afc90947ee320161bbf18eb6048`

It fails on the same block over and over, but every 10m or so it doesn't error right away and advances:
```
+10m: 0000000067217a46c49054bad67cda2da943607d326e89896786de10b07cb7c0
+20m: 0000000055fcaf04cb9a82bb86b46a21b15fcaa75ac8c18679b0234f79c4c615
+30m: 0000000000000591c56b14f82564036a1205b99e2831e2e89ea56eb75f165240
```
Thanks for your help!

_Originally posted by @RR82 in https://github.com/romanz/electrs/issues/308#issuecomment-841339030_

romanz:
Seems like a sporadic failures on various height blocks:
```
$ bitcoin-cli getblockheader 00000000839a8e6886ab5951d76f411475428afc90947ee320161bbf18eb6048 | jq .height
1
$ bitcoin-cli getblockheader 0000000067217a46c49054bad67cda2da943607d326e89896786de10b07cb7c0 | jq .height
2001
$ bitcoin-cli getblockheader 0000000055fcaf04cb9a82bb86b46a21b15fcaa75ac8c18679b0234f79c4c615 | jq .height
6001
[...]

PR #1291, opened 2026-05-21
Warn when bitcoind has not whitelisted electrs

### LINKED ISSUE #400: Detect p2p not being whitelisted
It'd be great if electrs at least detected this. Ideally fall back to RPC or some other technique. IDK if whitelist affects this setting but we should probably recommend whitelisting 127.0.0.1 just like NBXplorer does (it even detects lack of whitelisting and logs a message).

_Originally posted by @Kixunil in https://github.com/romanz/electrs/issues/399#issuecomment-843901638_

PeterXMR:
## Summary

Closes #400.

If electrs is not granted the `download` permission by bitcoind, `getheaders` is silently ignored during IBD and the connection can be dropped once `maxuploadtarget` is reached — both leading to opaque sync failures (see #399 for the original report).

This change makes electrs detect that condition at startup and emit a single clear warning, modeled on NBXplorer's behavior (referenced in the issue thread):

```
WARN electrs::daemon] electrs is not whitelisted by bitcoind: it may be
disconnected during IBD or once `maxuploadtarget` is reached. Consider
adding `whitelist=download@127.0.0.1` to bitcoin.conf. See
https://github.com/romanz/electrs/issues/400 for details.
```

### How it works

After the p2p version handshake completes, electrs calls `getpeerinfo`, finds its own peer entry by matching `addr` against the local socket address, and checks whether the `permissions` array contains `"download"`. Any failure mode (RPC error, no matching peer, missing field, address mismatch) is logged at `debug!` only and never affects startup.

The parsing logic is factored into a pure `check_whitelist_status(peers, local_addr)` helper for unit testing.

### Scope

Detection + warning only. The "ideally fall back to RPC" half of the original ask is out of scope here — that design space is already explored in #1274 (ZMQ+REST) and #1252 (bindex), so a third design in this PR would conflict.

## Test plan

- [x] 8 new unit tests for `check_whitelist_status` (download present/absent, no matching peer, missing `permissions` field, IPv4-mapped-IPv6 normalization, outbound-peer skip, multi-peer, empty list)
- [x] `cargo test --lib` — 29/29 pass
- [x] `cargo clippy --all-targets -- -D warnings` — clean
- [x] `cargo fmt --check` — clean
- [x] Live tested against Bitcoin Core v31.0.0 on signet:
  - Without whitelist → warning fires
  - With `whitelist=download@127.0.0.1` → no warning, startup proceeds normally
  - Verified `getpeerinfo[].addr` matches electrs's `local_addr` byte-for-byte (`127.0.0.1:<ephemeral>`) and `permissions: ["download"]` when whitelisted

### DIFF
--- a/src/daemon.rs
+++ b/src/daemon.rs
@@ -11,6 +11,7 @@ use serde_json::{json, value::RawValue, Value};
 
 use std::fs::File;
 use std::io::Read;
+use std::net::SocketAddr;
 use std::path::Path;
 
 use crate::{
@@ -139,11 +140,9 @@ impl Daemon {
             bail!("electrs requires non-pruned bitcoind node");
         }
 
-        let p2p = Mutex::new(Connection::connect(
-            config.daemon_p2p_addr,
-            metrics,
-            config.magic,
-        )?);
+        let p2p = Connection::connect(config.daemon_p2p_addr, metrics, config.magic)?;
+        warn_if_not_whitelisted(&rpc, p2p.local_addr());
+        let p2p = Mutex::new(p2p);
         Ok(Self { p2p, rpc })
     }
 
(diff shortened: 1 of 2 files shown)


romanz:
https://github.com/romanz/electrs/pull/1252 will fetch blocks using REST API instead of p2p.

Ari4ka (approved):

PR #1301, opened 2026-05-29
ci: integration test against bitcoin/bitcoin#35182 @ ddefe4263b

pinheadmz:
Replace the 31.0rc2 prebuilt-binary download with a docker stage that pulls pinheadmz/bitcoin:pr35182-ddefe4263b (multi-arch) and copies bitcoind/bitcoin-cli out of /opt/bin/, so the electrs integration container exercises bitcoin/bitcoin#35182 (HTTP server rewrite) at commit ddefe4263b42aa050a4bd55c073e1640f06e7332.

Drop the daily cron from rust.yml; we only need push/PR triggers here.

### DIFF
--- a/.github/workflows/rust.yml
+++ b/.github/workflows/rust.yml
@@ -4,9 +4,8 @@ on:
   push:
     branches:
       - master
+      - pr35182-*
   pull_request:
-  schedule:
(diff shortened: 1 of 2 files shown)


pinheadmz:
sorry wrong upstream

Issue #1239, opened 2025-10-27
Bug: crash on wallet connect followed by the never ending restart loop


mariodian:
**Describe the bug**
When I connect my wallet (Sparrow) to Electrs, it loads new transactions and gets into a restart loop shortly after. Sometimes it can go days without restarting, sometimes it happens within a few seconds of connecting with Sparrow. 

I then usually have to restart the machine to be able to run Electrs again without crashing on me immediately. It doesn't matter whether I run it from systemd or manually, the behavior is the same.

**/var/log/syslog**
```
2025-10-27T23:42:18.874823+08:00 nuc13 electrs[8474]: [2025-10-27T15:42:18.874Z INFO  electrs::chain] chain updated: tip=000000000000000000017a6ff3e5cb2a2652438aea07a478266d5899d9a16f67, height=921042
2025-10-27T23:45:55.600701+08:00 nuc13 electrs[8474]: [2025-10-27T15:45:55.600Z WARN  electrs::thread] p2p_send thread failed: p2p failed to send
2025-10-27T23:45:55.600781+08:00 nuc13 electrs[8474]: [2025-10-27T15:45:55.600Z WARN  electrs::thread] because: Connection reset by peer (os error 104)
2025-10-27T23:45:55.603475+08:00 nuc13 electrs[8474]: [2025-10-27T15:45:55.602Z WARN  electrs::electrum] RPC blockchain.scripthash.subscribe failed: failed to get block 00000000000000000002d57ffa33607da376a83fcc74377cff68d7ce5e81e7d0: receiving on an empty and disconnected channel
2025-10-27T23:45:55.603549+08:00 nuc13 electrs[8474]: thread 'main' panicked at src/electrum.rs:346:26:
2025-10-27T23:45:55.603579+08:00 nuc13 electrs[8474]: missing scripthash status
2025-10-27T23:45:55.603601+08:00 nuc13 electrs[8474]: note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace
2025-10-27T23:45:55.609837+08:00 nuc13 electrs[8474]: [2025-10-27T15:45:55.609Z INFO  electrs::db] closing DB at ./db/bitcoin
2025-10-27T23:45:55.914719+08:00 nuc13 systemd[1]: electrs.service: Main process exited, code=exited, status=101/n/a
2025-10-27T23:45:55.914815+08:00 nuc13 systemd[1]: electrs.service: Failed with result 'exit-code'.
[...]

PR #1324, opened 2026-07-29
fix(electrum): replace panic with error for missing scripthash status in subscription

mariodian:
I spoke to @Kixunil about this and he recommended I created a PR because we shouldn't panic here. I get these errors quite often with Sparrow and it tends to crash my electrs sometimes (#1239). Even with this fix, I get occasional crashes, but less frequent.

### DIFF
--- a/src/electrum.rs
+++ b/src/electrum.rs
@@ -367,7 +367,7 @@ impl Rpc {
                 Entry::Vacant(e) => {
                     let status = results
                         .remove(scripthash)
-                        .expect("missing scripthash status")?; // return an error for failed subscriptions
+                        .ok_or_else(|| anyhow::anyhow!("missing scripthash status"))??;
                     e.insert(status).statushash()
                 }
             };


src/electrum.rs
```diff
@@ -357,7 +367,7 @@ impl Rpc {
                 Entry::Vacant(e) => {
                     let status = results
                         .remove(scripthash)
-                        .expect("missing scripthash status")?; // return an error for failed subscriptions
+                        .ok_or_else(|| anyhow::anyhow!("missing scripthash status"))??;
```
Kixunil, line 370: FTR, I believe this happens when the wallet sends the registration for the same script hash twice in the same request.

Kixunil (approved):
This is a good way to avoid crashes but it's a good question if this should ever error. Maybe instead we should rewrite the code to accept duplicates and handle them soundly.

The entire function looks quite over-complicated to me but I guess it has something to do with parallel querying. (Is it even useful? Isn't daemon connection locked anyway?)

PR #1325, opened 2026-07-29
Update transaction.id_from_pos arg parsing for API compliance

jharveyb:
A follow-up to https://github.com/romanz/electrs/pull/1262 .

There are two non-compliant behaviors with transaction.id_from_pos in 0.11.1:

- The 3rd parameter, 'merkle' should be optional, not required. The server should treat it as false if not present.
- If 'merkle' is false, the response should not be a JSON object, but just the bare TXID in hex.

Source:

https://electrum-protocol.readthedocs.io/en/latest/protocol-methods.html#blockchain-transaction-id-from-pos

This behavior has not changed since version 1.4. Note, this is what the Blockstream electrs does right now:

https://github.com/Blockstream/electrs/blob/new-index/src/electrum/server.rs#L528

This PR follows the structure of the transaction.get optional `verbose` param handling. I can remove the tests if you like.

This may break the `rust-electrum-client` when `merkle=false` - I'll open a matching PR there.

Update, 31/07/26: PR on `rust-electrum-client` to add compatibility with electrs <=0.11.1:

https://github.com/bitcoindevkit/rust-electrum-client/pull/222

Would be cool to get this in for 0.12.0!

### DIFF
--- a/src/electrum.rs
+++ b/src/electrum.rs
@@ -79,6 +79,22 @@ impl From<&TxGetArgs> for (Txid, bool) {
 
 #[derive(Deserialize)]
 #[serde(untagged)]
+enum TxFromPosArgs {
+    HeightPos((usize, usize)),
+    HeightPosMerkle((usize, usize, bool)),
+}
+
+impl From<&TxFromPosArgs> for (usize, usize, bool) {
+    fn from(args: &TxFromPosArgs) -> Self {
+        match args {
+            TxFromPosArgs::HeightPos((height, tx_pos)) => (*height, *tx_pos, false),
+            TxFromPosArgs::HeightPosMerkle(args) => *args,
+        }
+    }
+}
+
+#[derive(Deserialize)]
+#[serde(untagged)]
 enum BroadcastArgs {
     Package((Vec<String>,)),
     PackageVerbose((Vec<String>, bool)),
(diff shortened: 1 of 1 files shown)


Juwon-Ogunseye:
looked through the transaction_from_pos diff, checks out. TxFromPosArgs enum defaults merkle to false when missing same as verbose does for TxGetArgs, and the json!(txid) change for merkle=false matches the protocol docs. anyone added tests for both the 2 and 3 arg cases yet?

jharveyb:
> looked through the transaction_from_pos diff, checks out. TxFromPosArgs enum defaults merkle to false when missing same as verbose does for TxGetArgs, and the json!(txid) change for merkle=false matches the protocol docs.

Thanks for taking a look!

> anyone added tests for both the 2 and 3 arg cases yet?

The current added tests check that both cases parse correctly:

https://github.com/romanz/electrs/pull/1325/changes#diff-ede2fa6030c1edb89d5de41ac20473c31669e46197974a46108a9bff88eebfaeR877

And I manually tested against a build of electrs from this branch. I don't see formal integration tests here beyond testing with electrumX, so I wasn't planning to add those without some larger discussion with the maintainer(s) first.

romanz (approved):
LGTM, thanks!

PR #1330, opened 2026-08-28
Create SECURITY.md

mumintey:
```shell

```

### DIFF
--- a/SECURITY.md
+++ b/SECURITY.md
@@ -1,0 +1,21 @@
+# Security Policy
+
+## Supported Versions
+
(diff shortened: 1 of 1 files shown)


Kixunil:
This is not helpful. A security policy is for @romanz to decide. You didn't even suggest what you think would make sense.

Personally, I think we should just recommend people connect to electrs via authenticated means and we should exclude DoS attacks from our threat model but include other attacks via RPC.

romanz:
Closing in favor of https://github.com/romanz/electrs/pull/1331.

PR #1332, opened 2026-08-30
Create SECURITY.md for security policy

mumintey:
Added a security policy document outlining supported versions and vulnerability reporting.


romanz:
See #1331.

PR #1335, opened 2026-08-31
docs: clarify script-hash status wording

ginavalent:
Clarify script-hash terminology in the status documentation so it matches electrs and Bitcoin RPC usage.


romanz (approved):
Thanks

